PR #37, opened 2020-01-02
Adding minor changes to simplify testing

eginez:
Hey guys this is a set of minimal changes to the command line runner script to make it friendlier to our testing environment. 

Any thoughts on making this part of master so we don't have to keep it  in a fork?

### DIFF
--- a/build-feature.sh
+++ b/build-feature.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 cd spring-graal-native-feature
-./mvnw clean package
+mvn clean package
--- a/spring-graal-native-samples/commandlinerunner/compile.sh
+++ b/spring-graal-native-samples/commandlinerunner/compile.sh
@@ -13,7 +13,7 @@ rm -rf target
 mkdir -p target/native-image
 
 echo "Packaging $ARTIFACT with Maven"
-../../mvnw -DskipTests package > target/native-image/output.txt
+mvn -DskipTests package > target/native-image/output.txt
 
 JAR="$ARTIFACT-$VERSION.jar"
 rm -f $ARTIFACT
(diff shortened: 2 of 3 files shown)


build-feature.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 cd spring-graal-native-feature
-./mvnw clean package
```
sdeleuze: Using the Maven wrapper is a best practice that allows to build the project on environments without Maven installed, could you elaborate on the reasons why you need these changes?

spring-graal-native-samples/commandlinerunner/compile.sh
```diff
@@ -28,14 +28,15 @@ CP=BOOT-INF/classes:$LIBPATH:$FEATURE
...
 time native-image \
+  --verbose \
   --no-server \
   --no-fallback \
   -H:Name=$ARTIFACT \
   -H:+ReportExceptionStackTraces \
   --allow-incomplete-classpath \
   --report-unsupported-elements-at-runtime \
   -DremoveUnusedAutoconfig=true \
-  -cp $CP $MAINCLASS >> output.txt
```
sdeleuze: By default I think we would like to not "pollute" the output with too many information, could you please elaborate on your need?

build-feature.sh
eginez: Well this downloads a file from the internet and writes to a particular location in the disk, in some environments that is no feasible

spring-graal-native-samples/commandlinerunner/compile.sh
eginez: You are right, I am already outputting the content in case of a failure this is probably not needed

build-feature.sh
sdeleuze: OK I understand.

sdeleuze:
Merge via 68e48c395fa942815a02e012de91bb834f1331d2 and applied to other samples as well.

PR #50, opened 2020-02-20
Spring Feature support for Apache Tomcat

fhanik:


fhanik:
Do not merge yet, bundles may have to be conditional based on classpath search

PR #57, opened 2020-03-25
Add WebClient sample

dsyer:
Currently broken in native image build.


sbrannen:
What exactly is broken?

Do you get a stack trace, error message, etc.?

dsyer:
[output.txt](https://github.com/spring-projects-experimental/spring-graal-native/files/4381860/output.txt)

sbrannen:
OK, so the following is the core issue here.

```
Error: Classes that should be initialized at run time got initialized during image building:
 io.netty.util.NetUtil the class was requested to be initialized at build time (from feature org.springframework.graal.support.InitializationHandler.lambda$register$0). To see why io.netty.util.NetUtil got initialized use -H:+TraceClassInitialization
io.netty.util.AbstractReferenceCounted the class was requested to be initialized at build time (from the command line). To see why io.netty.util.AbstractReferenceCounted got initialized use -H:+TraceClassInitialization
io.netty.channel.DefaultFileRegion the class was requested to be initialized at build time (subtype of io.netty.util.AbstractReferenceCounted). To see why io.netty.channel.DefaultFileRegion got initialized use -H:+TraceClassInitialization
```

Given "the class was requested to be initialized at build time (from feature org.springframework.graal.support.InitializationHandler.lambda$register$0)", @aclement, is this something you can address in the feature?

aclement:
I played with the branch. Got an error, fixed it up with a hint - but it wasn't that error... I can't seem to get the init error with either Java8 or Java11.  But yes, @sbrannen , if it were an init error like that we'd look to add it into `initialization.json` in the feature (after quickly confirming it was the fix by probably playing with it in the `compile.sh` script). I'd be a little wary here as it is a netty one and netty is one of the few shipping their own config - so I might quickly dig into their jar to see what config they are providing and whether we would be trying to contradict it. Maybe I need to try it on linux and see what error I get.

PR #125, opened 2020-05-15
Add devcontainers to actuator-webmvc

dsyer:


sdeleuze:
Replaced by #154 which will have a wider scope.

PR #132, opened 2020-05-20
Exclude Spring Data Web configuration in PetClinic

dsyer:
Can be reverted when Spring Data is fixed to add
proxyBeanMethods = false


sdeleuze:
Merge via 9f9ea86756f738ef77cab31cff32f8fe9a3e3f66 thanks.

PR #139, opened 2020-05-24
Correct option for dumpConfig

jkroepke:
I tried so debug my program as described in the documentation but after a while I realize that the dumpConfig option is wrong.


sdeleuze:
Merged, thanks!

PR #140, opened 2020-05-24
reflectCompare.sh: org.springframework.graal.support -> org.springframework.graalvm.support

jkroepke:
Looks like the scripts are forgotten on rename.


sdeleuze:
Good catch, thanks.

PR #170, opened 2020-06-09
Fix the webflux-r2dbc sample

christophstrobl:
Do not merge - needs clarification for trigger h2 setup!

The first of the two commits fixes the issue with the `ReactiveCrudRepository` not being picked up due to failing checks for reactive adapters in the `ReactiveAdapterRegistry`, which is checking for `Flux` where the `Hint` for _r2dbc` only provides `Mono`.

The 2nd commit adds hints for setting up the required embedded h2 instance. All the changes in here should move to a dedicated place and not reside in the `Hints` for r2dbc.


sdeleuze:
Merged and polished via https://github.com/spring-projects-experimental/spring-graalvm-native/commit/9f5208bb6af15c4105f021848e6b1364d7ff4740. Thanks!

Issue #158, opened 2020-06-05
Improve SpringData custom repository implementation support.


christophstrobl:
Lookup and register custom implementations of SpringData repository fragments.

```java
interface OrderRepository extends CrudRepository<Order, String>, OrderRepositoryCustom {
  // ..
}

@Component
class OrderRepositoryImpl implements OrderRepositoryCustom {
  // ..
}
```

Investigate how far it is possible to move away from the default `Impl` postfix towards reading the `repositoryImplementationPostfix` value if set.

Check if it is required to have the `@Component` annotation on the actual custom implementation or if there's a way around it.

sdeleuze:
Rollbacked due to error on Java 11 builds (on [20.1](https://ci.spring.io/teams/spring-graalvm-native/pipelines/spring-graalvm-native/jobs/20.1-dev-java11-samples-build/builds/79) and [master](https://ci.spring.io/teams/spring-graalvm-native/pipelines/spring-graalvm-native/jobs/master-java11-samples-build/builds/34)).

PR #175, opened 2020-06-10
Refine SpringDataComponentProcessor.

christophstrobl:
The `SpringDataComponentProcessor` now detects required annotations and registers them for reflective access so they do no longer need to be part of the `Hints`.
Some of the logic held in `ResourceHandler` moved to the `ComponentProcessor`.
Logging has also been improved to be less verbose by default.

Closes: #158


aclement:
It looks ok to me - the need for `hasReflectionConfigFor(String key);` I'd like to review in more detail when there is time but I'm glad to see some of the hacked up code in resource handler get removed and handled properly in the component processor. Is there still a need for `processRepository2()` in the resources handler? Nice to see some tidy up to lazily resolve things in the Type object. All forward progress, thanks!

PR #277, opened 2020-09-10
Add Spring Data Elasticsearch hints and sample

christophstrobl:
Provide native image hints that add required proxy and reflection for Spring Data Elasticsearch Repositories.

Provide a `Map` based substitution for the `org.apache.http.impl.client.BasicAuthCache` used by the Elasticsearch `RestClient`.

Move Reactive Spring Data hints to dedicated class, so they can be used by `TypeInfo` imports.

Update docker configuration to include and start Elasticsearch server.


christophstrobl:
@sdeleuze Please have a look at the Elasticsearch setup in the Docker and build files. Thank you!

spring-graalvm-native-configuration/src/main/java/org/springframework/boot/autoconfigure/data/RepositoriesHints.java
```diff
@@ -20,34 +20,43 @@
...
 
-@NativeImageHint(trigger = CassandraReactiveRepositoriesAutoConfiguration.class, typeInfos= {
-		@TypeInfo(types= {ConditionalOnRepositoryType.class, OnRepositoryTypeCondition.class,RepositoryType.class})
+@NativeImageHint(trigger = CassandraReactiveRepositoriesAutoConfiguration.class, typeInfos = {
+		@TypeInfo(types = {ConditionalOnRepositoryType.class, OnRepositoryTypeCondition.class, RepositoryType.class})
 })
-@NativeImageHint(trigger = CassandraRepositoriesAutoConfiguration.class, typeInfos= {
-		@TypeInfo(types= {ConditionalOnRepositoryType.class, OnRepositoryTypeCondition.class,RepositoryType.class})
+@NativeImageHint(trigger = CassandraRepositoriesAutoConfiguration.class, typeInfos = {
+		@TypeInfo(types = {ConditionalOnRepositoryType.class, OnRepositoryTypeCondition.class, RepositoryType.class})
```
aclement, line 36: I know this is more of a reformatting of what is there than a real change but it caught my eye that I think the feature should be auto adding at least the first two of these, we should try testing with the first removed and then the second. (same applies across all these hints)

christophstrobl: thanks @aclement - checked without the hints and everything worked. Will undo the changes and open another PR that takes care of removing hints for the other store modules.

PR #287, opened 2020-09-17
Add Hints and example for Data Neo4j.

michael-simons:
This PR provides `Neo4jRepositoriesHints` and `Neo4jReactiveRepositoriesHints`, triggered by the appropriate auto configuration classes in the latest Spring Boot 2.4.
The hints have been added to `RepositoriesHints`.

In addition, `data-neo4j` as an examples has been added. It's basically a copy of the `data-mongo` sample, with the same domain and the same queries, but for Neo4j.

For a successful run of `build.sh`, a locally running Neo4j instance is required. The easiest way to get one is

```
docker run --publish=7474:7474 --publish=7687:7687 -e 'NEO4J_AUTH=neo4j/secret' neo4j:4.1
```

Please feel free to adapt as necessary. 

cc @christophstrobl @meistermeier


sdeleuze:
Thanks for this contribution @michael-simons. We currently can't use Docker on Docker, so we would need to install Neo4j on our `spring-graalvm-native` Docker image which will be not straightforward because Neo4j requires requires Java 11+. Since it will require some work on that side I move it to 0.8.2.

michael-simons:
I think using `neo4j:3.5` (requiring JDK 8) should work equally well. I need to add in some config for the connection to work, but that's easily doable.

How are you running MongoDB or elastic? Did you install those on the base image? I cannot imagine you're running their products natively… If so, we would like to have your help making Neo4j native runnable, too :D

michael-simons:
So, I switched the compose script to Neo4j 3.5 and added a hint to the pom about the need for all security services when people might want to use an encrypted connection. Please let me know if this works for you.

sdeleuze:
Nice, we can maybe try to make it runnable for 0.8.1 then.

Do you want to try to add support yourself based on what we did for elasticsearch in https://github.com/spring-projects-experimental/spring-graalvm-native/commit/fe0e651de4890708bd814deeda87dc1b3e095dc4 and https://repo.spring.io/milestone/org/springframework/experimental/spring-graalvm-native-docs/0.8.0/spring-graalvm-native-docs-0.8.0.zip!/reference/index.html#_using_container_based_build_environment ?

michael-simons:
Hi, I tried but I failed miserably. Our deb package is complaining about openjdk not installed (and also about package `daemon`). The later is fixable of course, but it seems not reasonable to replace the jdk.

I started the CI container with `docker run -it spring-graalvm-native-dev:20.2-dev-java8  /bin/bash` to run my experiments.

When I fix the daemon package (with `apt-get install daemon`), and than do the following:

```
curl -LO https://dist.neo4j.org/deb/neo4j_3.5.21_all.deb \
 && dpkg  --force-all -i neo4j_3.5.21_all.deb
```

In the container, I end up with 

```
Setting up neo4j (1:3.5.21) ...
invoke-rc.d: could not determine current runlevel
invoke-rc.d: policy-rc.d denied execution of start.
```

Here are our instructions

https://neo4j.com/docs/operations-manual/3.5/installation/linux/debian/

The dl link to the deb is above, tarball would be https://neo4j.com/artifact.php?name=neo4j-community-3.5.22-unix.tar.gz

If you can help me getting the service installed, I can add the missing config bits.

michael-simons:
I managed to add a running Neo4j 3.5 to the container (also removed the tons of elastic search stack traces), but I cannot manage to get anything build in either the dev or master image. However, mongo or elastic fail as well. I guess this is unrelated.

sdeleuze:
Ok thanks I will have a deeper look.

sdeleuze:
I will take care of merging this PR.

sdeleuze:
Merged with a remaining issue (#300) which is maybe a regression on our side. Thanks a lot @michael-simons, great work.

Issue #307, opened 2020-09-25
Make webmvc-functional sample work with @SpringBootApplication


dsyer:
Currently it uses `@ImportAutoConfiguration` to manually import a selected set of configs, but breaks on a full-bore auto configuration (the webflux version of the same sample would work fine).

dsyer:
I found I needed this (plus the custom binder and remove request mapping config at build time):

```json
[
	{
		"name": "org.springframework.http.converter.FormHttpMessageConverter",
		"fields": [
			{
				"name": "partConverters"
			}
		]
	},
	{
		"name": "org.springframework.boot.web.servlet.server.Encoding",
		"allDeclaredConstructors": true,
		"allDeclaredMethods": true
	}
]
```

dsyer:
There's still an issue with the /error endpoint in functional mode even with the above JSON.

```
2020-09-25 07:27:24.334  WARN 1 --- [nio-8080-exec-1] o.s.web.servlet.PageNotFound             : No mapping for GET /error
```

dsyer:
See also: https://github.com/spring-projects/spring-boot/issues/23485

dsyer:
Fixed by #311

PR #311, opened 2020-09-28
Update webmvc-functional to work with @SpringBootApplication

dsyer:
Works now, but maybe wait for Spring Boot issue to be fixed
before merging (and take out the JSON).


dsyer:
Fixes #307

Issue #289, opened 2020-09-17
Security example with @EnableGlobalMethodSecurity  causes  java.io.FileNotFoundException 


efrengarcial:
hi,

I have updated the "Security" example and added the annotation @EnableGlobalMethodSecurity  to the class WebSecurityConfig:

Sample Project: spring-graalvm-native-samples/security

```@Configuration(proxyBeanMethods = false)
@EnableWebSecurity
@EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true)
public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
	@Override
	protected void configure(HttpSecurity http) throws Exception {
		http
			.authorizeRequests()
				.antMatchers("/", "/home").permitAll()
				.anyRequest().authenticated()
				.and()
			.formLogin()
				.loginPage("/login")
				.permitAll()
				.and()
			.logout()
				.permitAll();
	}

	@Bean
	@Override
	public UserDetailsService userDetailsService() {
		UserDetails user =
			 User.withDefaultPasswordEncoder()
				.username("user")
				.password("password")
				.roles("USER")
				.build();

		return new InMemoryUserDetailsManager(user);
	}
}
```

But it fails at the moment of running the application, Full Stack Trace:

```
2020-09-17 08:02:30.766  INFO 11442 --- [           main] c.e.securingweb.SecuringWebApplication   : Starting SecuringWebApplication using Java 11.0.8 on server with PID 11442 (/opt/java/apps/spring-graalvm-native/spring-graalvm-native-samples/security/target/security started by efren in /opt/java/apps/spring-graalvm-native/spring-graalvm-native-samples/security)
2020-09-17 08:02:30.767  INFO 11442 --- [           main] c.e.securingweb.SecuringWebApplication   : No active profile set, falling back to default profiles: default
2020-09-17 08:02:30.790  WARN 11442 --- [           main] ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanDefinitionStoreException: Failed to process import candidates for configuration class [com.example.securingweb.WebSecurityConfig]; nested exception is java.io.FileNotFoundException: class path resource [org/springframework/security/config/annotation/method/configuration/MethodSecurityMetadataSourceAdvisorRegistrar.class] cannot be opened because it does not exist
[...]

PR #364, opened 2020-11-17
Support Pre and Post method security annotations

eleftherias:
See gh-289

### DIFF
--- a/spring-graalvm-native-configuration/src/main/java/org/springframework/PrePostSecuredComponentProcessor.java
+++ b/spring-graalvm-native-configuration/src/main/java/org/springframework/PrePostSecuredComponentProcessor.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework;
+
+import org.springframework.graalvm.extension.ComponentProcessor;
+import org.springframework.graalvm.extension.NativeImageContext;
+import org.springframework.graalvm.type.Type;
+
+import java.util.ArrayList;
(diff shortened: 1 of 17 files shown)


spring-graalvm-native-configuration/src/main/java/org/springframework/PrePostSecuredComponentProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Recognize spring.components that use any of @PostAuthorize, @PostFilter, @PreAuthorize or @PreFilter
+ * annotation and register proxies for them.
+ *
+ * @author Eleftheria Stein
+ */
+public class PrePostSecuredComponentProcessor implements ComponentProcessor {
```
eleftherias, line 31: Is there a way to have this `ComponentProcessor` only trigger when `@EnableGlobalMethodSecurity` is present?

aclement: Do I presume that `@EnableGlobalMethodSecurity` may just be 'anywhere in the app'? (Excuse my lack of spring security knowledge) and so that is why we can't easily check for it in the handle() method because it may be on a component you haven't looked at yet or one you saw earlier and we aren't remembering. Currently there is no control for turning off component processors like that, but i'm sure we could work something out.

eleftherias: Yes, that's right.
This may be a specific use case, since `@PreAuthorize` can't be used unless `@EnableGlobalMethodSecurity` is present.
I was thinking there might be a way to specify processors in`@NativeImageHint` based on a `trigger`.
The goal would be to avoid taking the time to process each component, when we already know that none fit the criteria in `handle`.

sdeleuze: @eleftherias Should I merge this PR as it is to be part of 0.8.3, or should we wait it to evolve based on your discussion with @aclement ?

eleftherias: @sdeleuze I think it can be merged and we can make the optimisations later.

aclement: sorry for my slow reply. I think we can build something to optimize this yes. I think most solutions come down to some kind of pre-pass of components so any global context can be determined, before running the processors against each individual component. I bet there are a number of use cases that need this kind of thing. Will think about the best thing to do here.

Issue #354, opened 2020-11-13
Add Spring Cloud Config support


aclement:
Currently it runs with the agent generated config. We should eliminate that, capturing it as hints/smarts in the feature/configuration modules.

PR #501, opened 2021-02-24
Remove json files for config-client-new

spencergibb:
Fixes gh-354

/cc @ryanjbaxter @aclement


sdeleuze:
I don't see additional hints in this PR, maybe you have them locally?

spencergibb:
So far, for config client new it didn't need them. Config client likely will.

sdeleuze:
I did a quick test, with the native config `config-client-new` is green, if I remove the `META-INF/native-image` directory and then run `build-without-server-build.sh` it fails. Could you please double check?

spencergibb:
You have to do `./build.sh` the first time. Only then, can you call `build-without-server-build.sh`

sdeleuze:
Yeah that's what I did and `build-without-server-build.sh` failed.

spencergibb:
You'll notice I also made a change to the sample app itself

aclement:
I've fixed the problem that necessitated changing the client app - either should work (separate commandlinerunner or simple bean). And can confirm deleting the config for the '-new' variant leaves you with a working app now.

spencergibb:
Closing due to the move of the samples directory.

Issue #514, opened 2021-02-25
Image build fails on Windows seemingly because of backslashes in spring-aot\resource-config.json


leccelecce:
[repro-sn514.zip](https://github.com/spring-projects-experimental/spring-native/files/6043154/repro-sn514.zip)
0.9.0-SNAPSHOT. My dev environment is Windows 10.

Build fails in the attached repro project with the below exception:
```
[INFO]     [creator]     Fatal error:java.util.regex.PatternSyntaxException: Illegal/unsupported escape sequence near index 7
[INFO]     [creator]     static\images\test.png
[INFO]     [creator]            ^
[INFO]     [creator]     	at java.base/java.util.regex.Pattern.error(Pattern.java:2027)
```

I believe this is because of the backslashes in this file:
 `target\classes\META-INF\native-image\org.springframework.aot\spring-aot\resource-config.json`:

![image](https://user-images.githubusercontent.com/24962424/109158231-fffde480-776a-11eb-8bb4-fb0c0c62b434.png)

I'm not actually sure if this line is even required, because further up in the file there is a more generic regex:

![image](https://user-images.githubusercontent.com/24962424/109158419-3cc9db80-776b-11eb-99fb-26dc3771c0e3.png)

Also noted that the resource-config file exists in two locations, target\generated-sources and also target\classes

[edit]
Not 100% sure of myself here because if I manually delete the offending lines from the two resource-config.json lines (in between them being generated and the image being built) I still seem to get the error. I'm not sure if they've already been packaged up into the JAR file by that point and so get read out of there again later though.

[edit-2]
This is definitely the source of the problem - I was able to set a breakpoint on BootstrapCodeGenerator immediately after `resource-config.json` has been written, allowing me time to go in and updated the two files to replace \\ with /, then allowing the Maven run to continue successfully. So again, I think this is a breaks-on-windows issue.

Possibly a similar issue to #507

sdeleuze:
Duplicate of #507.

sdeleuze:
OK, should be fixed by cef5c60ad3fb8fe4bfdfc899513615234a3e49e0.

PR #520, opened 2021-02-25
Fix build failure on Windows due to assuming forward slash separator

leccelecce:
Closes gh-514


leccelecce:
Closed this one as although it creates a working result, it was resulting in double forward slashes in resource-config.json which are incorrect.

PR #527, opened 2021-02-26
Switch to one sentence per line

snicoll:
This commit updates the reference guide to use one sentence per line.A
number of dead links have been fixed and additional attributes have
been introduced to avoid the duplication.


sdeleuze:
Merged, thanks!

Issue #355, opened 2020-11-13
Add Spring Cloud Discovery support


aclement:
Currently it runs with the agent generated config. We should eliminate that, capturing it as hints/smarts in the feature/configuration modules.

PR #530, opened 2021-02-26
Removes native json hints

spencergibb:
Adds ConfigServerHints

Initial TypeHint is for jgit.


sdeleuze:
Merged via 4d5d112d5ec337a34544f3a5b360aea1ce7661d9 with an accidental additional removal of Discovery Client native configuration. Could you please add Discovery Client hints in order to make it possible to close #355?

PR #521, opened 2021-02-25
Fix build failure on Windows due to assuming forward slash separator

leccelecce:
Closes gh-514

PR #535, opened 2021-03-01
Fix resource backslash issue at source rather than during JSONify

leccelecce:
This is an improved version of previously merged PR https://github.com/spring-projects-experimental/spring-native/pull/521 which was merged in https://github.com/spring-projects-experimental/spring-native/commit/cef5c60ad3fb8fe4bfdfc899513615234a3e49e0

The original fix just fixed the Windows backslash paths to forward-slashes during the final JSONify step in `ResourcesJsonMarshaller`. I went back and had a look at where the original issue stems from in `BootstrapCodeGenerator` and fixed it there instead (which was another use of Files.walk()).

Again, tested on Windows but would be good to have a check on a Linux system.

@aclement, I have not touched a pre-existing `String.replace("\\/","/")`  in `ResourcesJsonMarshaller` which appears to be from an earlier version of the code as I'm not sure what this is trying to do - if you have any recollection possibly this could be moved as well?

I did also consider making a change in `ResourcesDescriptor.add()` method but this seemed like a better approach.

### DIFF
--- a/spring-aot/src/main/java/org/springframework/aot/BootstrapCodeGenerator.java
+++ b/spring-aot/src/main/java/org/springframework/aot/BootstrapCodeGenerator.java
@@ -83,15 +83,16 @@ public class BootstrapCodeGenerator {
 				if (Files.exists(resourceFolder)) {
 					Files.walk(resourceFolder).filter(p -> !p.toFile().isDirectory()).forEach(p -> {
 						String resourcePattern = p.toString().substring(resourceFolderLen);
-						if (!resourcePattern.startsWith("META-INF/native-image")) {
+						String platformNormalisedResourcePattern = resourcePattern.replace("\\", "/");
+						if (!platformNormalisedResourcePattern.startsWith("META-INF/native-image")) {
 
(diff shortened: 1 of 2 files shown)


leccelecce:
Noticed that the existing check for `startsWith("META-INF/native-image")` was always broken on Windows as the pattern would never have a forward-slash - moved the String.replace() call above the if statement to resolve.

leccelecce:
I have rebased this on top of my other PR https://github.com/spring-projects-experimental/spring-native/pull/546 for now given the other one is not awaiting additional review. This should not be merged until PR https://github.com/spring-projects-experimental/spring-native/pull/546 has been to avoid conflicts.

sdeleuze:
@aclement Feel free to merge this one if you think that's relevant.

aclement:
Seems to behave on linux, thanks for the contribution!

sdeleuze:
@leccelecce Could you please sign [Spring Contributor License Agreement](https://cla.pivotal.io/sign/spring) if not already done?

Issue #524, opened 2021-02-25
Duplicate entries in spring-aot generated resource-config.json


leccelecce:
This is a relatively minor issue as far as I can tell. My resource-config.json generated by spring-aot contains both wildcarded directories, and the complete listing of the directory contents as well. Presumably only one is necessary.

Near the top:
```
      {
        "pattern": "^static/.*"
      },
      {
        "pattern": "^templates/.*"
      }
```

Further down:

```
      {
        "pattern": "static/favicon.ico"
      },
      {
        "pattern": "static/font/roboto/Roboto-Bold.eot"
      },
      {
        "pattern": "static/font/roboto/Roboto-Bold.ttf"
      },
      {
        "pattern": "templates/admin/admin.html"
      },
      {
        "pattern": "templates/admin/cache.html"
      },
```

leccelecce:
The actual regex values (i.e. `^static/.*` and `^templates/.*`) come from a `@ResourceHint` on `CommonWebInfos` (and there will be other equivalents elsewhere presumably).

The individual entries come from `BootstrapCodeGenerator.generate()` which seems to simply add everything present in the source resource folders.

```java
		if (!resourceFolders.isEmpty()) {
			logger.debug("Processing resource folders: " + resourceFolders);
			for (Path resourceFolder : resourceFolders) {
				int resourceFolderLen = resourceFolder.toString().length() + 1;
				if (Files.exists(resourceFolder)) {
					Files.walk(resourceFolder).filter(p -> !p.toFile().isDirectory()).forEach(p -> {
						String resourcePattern = p.toString().substring(resourceFolderLen);
						if (!resourcePattern.startsWith("META-INF/native-image")) {
							String platformNormalisedResourcePattern = resourcePattern.replace("\\", "/");
							logger.debug("Resource pattern: " + platformNormalisedResourcePattern);
							// TODO recognize resource bundles?
							// TODO escape the patterns (add leading trailing Q and E sequences...)
							buildContext.describeResources(crd -> crd.add(platformNormalisedResourcePattern));
						}
					});
				}
			}
		}
```

This is interesting as it's not going to be straightforward to test every resource added through this method against every existing pattern that may or may not be a regex - that feels very CPU intensive. I might put up a PR with a simple regex check for now.

PR #546, opened 2021-03-02
Reduce duplication of resources with basic regex cache

leccelecce:
Fixes https://github.com/spring-projects-experimental/spring-native/issues/524

This is a fairly basic check to avoid resources being duplicated twice in `resource-config.json` - once through a generic regex such as `"^templates/.*"` and again through an individual entry for every file in src/main/resources. The approach is to build a compiled regex cache of any obvious patterns built in the initial contributor step, and test files against those. The logic to identify an actual regex pattern is fairly basic (`patt.startsWith("^") && patt.endsWith(".*")`) but worked well for my semi-complex project. It looks like only `CommonWebInfos` uses this approach at the moment but rather than just hardcode checks for templates/ and static/ I thought this approach might be a bit more future proof.

Happy to consider an alternative approach.


aclement:
seems a reasonable change to me.

PR #552, opened 2021-03-03
Clean javadoc generation process

snicoll:
This commit moves Javadoc generation in the docs module itself, avoiding
a link to a parent directory. As a result, it is no longer necessary to
invoke the `javadoc` goal manually as it is triggered automatically
when the "docs" profile is active


sdeleuze:
Thanks!

PR #555, opened 2021-03-03
Discovery client hints

spencergibb:


samples/cloud-discovery/discoveryclient/src/main/resources/META-INF/native-image/reflect-config.json
```diff
@@ -0,0 +1,1274 @@
+[
+  { "name":"jdk.vm.ci.meta.JavaKind$FormatWithToString[]" },
+  { "name":"java.lang.reflect.AnnotatedElement[]" },
+  { "name":"java.lang.reflect.GenericDeclaration[]" },
```
spencergibb, line 4: I'm going to need some advice on where I should put these `java.lang` and `jdk.vm` entries should do.

sdeleuze: I think they should go in the hints that require them (`DiscoveryClientHints` like entries)

sdeleuze: Also `jdk.vm.ci.meta.JavaKind$FormatWithToString[]` looks suspicious, maybe try without that one.

spencergibb: fails without it
```
	at com.example.discoveryclient.DiscoveryClientApplication.main(DiscoveryClientApplication.java:21) ~[discoveryclient:0.0.1-SNAPSHOT]
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.cloud.loadbalancer.support.LoadBalancerClientFactory]: Factory method 'loadBalancerClientFactory' threw exception; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'default.org.springframework.cloud.loadbalancer.config.LoadBalancerAutoConfiguration.LoadBalancerClientSpecification': Unsatisfied dependency expressed through constructor parameter 1: Could not convert argument value of type [[Ljava.lang.String;] to required type [[Ljava.lang.Class;]: Failed to convert value of type 'java.lang.String[]' to required type 'java.lang.Class[]'; nested exception is java.lang.IllegalArgumentException: Class jdk.vm.ci.meta.JavaKind$FormatWithToString[] is instantiated reflectively but was never registered. Register the class by using org.graalvm.nativeimage.hosted.RuntimeReflection
```

aclement: I feel these hints probably relate to missing config around the conversion service. you are exercising some conversion nothing else has so far - on one hand that makes this something to attach to conversion service. But it would be good to understand precisely why it is there at some point (is there a different, better, conversion service hint that would mean we don't need this one)

aclement: The others, I'd probably start with discovery client (so attach them to the use case that needs them) but I bet when we see them cropping up in other situations we should gradually be able to steer to what the common root need is, and attach them to that. (It might be useful at some point to dump out which hints have overlapping configuration to rationalize a bit more. )

spencergibb: They ended up in `LoadBalancerHints` since that is what failed without them.

spencergibb:
```
  { "name":"org.springframework.cloud.client.ConditionalOnReactiveDiscoveryEnabled", "allDeclaredMethods":true }
```

Is the only one I don't know how to deal with. If I remove it, auto configuration classes are being loaded that have this as a guard so they shouldn't load.

It has a meta annotation of 
```
@ConditionalOnClass(name = "org.springframework.web.reactive.function.client.WebClient")
```

samples/cloud-discovery/discoveryclient/src/main/resources/META-INF/native-image/resource-config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "resources":{
+    "includes":[
+      {"pattern":"\\Qorg/springframework/cloud/netflix/eureka/reactive/EurekaReactiveDiscoveryClientConfiguration.class\\E"}
```
spencergibb, line 4: I'm unable to find a hint that allows this to be removed.

samples/cloud-discovery/discoveryclient/src/main/resources/META-INF/native-image/reflect-config.json
```diff
@@ -0,0 +1,3 @@
+[
+  { "name":"org.springframework.cloud.client.ConditionalOnReactiveDiscoveryEnabled", "allDeclaredMethods":true }
```
spencergibb, line 2: I'm unable to find a hint that allows this to be removed.

spencergibb:
I'm at the limits of what I can do with those last two entries.

sdeleuze:
Merged and polished via 8e97e999767c570c59dfb3901a8376c9a1484ac8.

PR #557, opened 2021-03-04
Introduces profiles for building with different databases.

schauder:
To run the build with hsqldb:
```
export DB=hsqldb
./build.sh
```


sdeleuze:
@schauder Could you please update the related `build.sh` in one of the data samples to test the drivers supported by those profiles?

schauder:
So It should run the test for all available profiles?

sdeleuze:
Yes I think so.

schauder:
> @schauder Could you please update the related `build.sh` in one of the data samples to test the drivers supported by those profiles?

Done.

PR #559, opened 2021-03-04
Restructure reference guide

snicoll:
This restructures the reference guide to harmonize it with other projects. The HTML rendering is now in `/reference/htmlsingle` while the pdf is in `/reference/pdf`. The PDF file was also renamed from `index.pdf` to `spring-native-reference.pdf`.

Various links to the doc were also updated as part of this change.


sdeleuze:
Thanks a lot Stéphane.

PR #587, opened 2021-03-11
 DB_HOST variable to connect to mySQL

jeevanjoseph:
This PR add a DB_HOST env var so that the location of the MySQL host can be changed easily. 

Today the app tries to connect to `localhost` always and the docker compose provides mysql at `mysql-server:3306`

using localhost will work for MySQL when 
- the mysql container is running 
- ports are mapped to  localhost
- the sample is run as a local process (not a container) - the `build.sh` seems to do this. 

If this PR is acceptable, I can update the other samples as well.

### DIFF
--- a/samples/petclinic-jpa/docker-compose.yml
+++ b/samples/petclinic-jpa/docker-compose.yml
@@ -14,6 +14,8 @@ services:
       - ./src/main/resources/db/mysql/data.sql:/docker-entrypoint-initdb.d/2-data.sql
   petclinic-jpa:
     image: petclinic-jpa:0.0.1-SNAPSHOT
+    environment:
+      - DB_HOST=mysql-server
     ports:
       - "8080:8080"
     depends_on:
--- a/samples/petclinic-jpa/src/main/resources/application.properties
+++ b/samples/petclinic-jpa/src/main/resources/application.properties
@@ -1,6 +1,6 @@
 # database init, supports mysql too
 database=mysql
-spring.datasource.url=jdbc:mysql://localhost/petclinic?serverTimezone=UTC&useUnicode=yes&characterEncoding=UTF-8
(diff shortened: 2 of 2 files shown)


sdeleuze:
@schauder Any advice on this PR?

pivotal-issuemaster:
@jeevanjoseph Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects-experimental/spring-native&pullRequestId=587) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

schauder:
I'm a little torn with this one.

I'd like to execute the petclinic with different databases, but I think we should integrate that with actually setting up the database using docker.

In any case we should keep the mechanism symmetric to the data-jpa example which currently uses profiles to control the database and is planned to extend that to other databases using docker. See https://github.com/spring-projects-experimental/spring-native/commit/fb40830f1cae0ae95d618836a95dd4219ce7faef

sdeleuze:
@jeevanjoseph Thanks for the PR, but I tend to agree with @schauder, it is probably better to deal with database consistently via containers for our samples.

jeevanjoseph:
@sdeleuze @schauder - Thanks for the feedback and the profiles are the better way to switch databases. However this PR was to remove the hardcoded “localhost” reference in the JDBC url regardless of the profile used. Currently both the pet clinic samples won’t work using the pure docker compose (running a `docker-compose-up`. App would come up, but db connections would fail)


The hardcoded `localhost` in the app properties (regardless of the profile used) would not work with containers since the app container and the db container would be different, and they cannot talk to each other over `localhost`. To be more specific , this would not work in `docker` or `docker-compose`. It would work on kubernetes if both containers are defined within the same PodSpec(or when using `podman`).

PR #591, opened 2021-03-11
Use version placeholder in documentation URL

snicoll:


sdeleuze:
I think I will decline this one since I prefer to link `current` like we do on start.spring.io since that's the link we want users and search engines to use.

snicoll:
Ah, right. Fair enough!

PR #592, opened 2021-03-11
Start building against Spring Boot 2.4.4-SNAPSHOT

snicoll:
I haven't executed the samples so I don't know if that'll break that part of the build or not.


sdeleuze:
Merged a polished commit via e9a346bdda626655bc12fc5a62f4b17bd9111940, thanks.

PR #632, opened 2021-03-18
Polish ci

snicoll:
This PR harmonizes the main build to use a custom CI image that can be reused for other tasks  (i.e. releasing).

A new CI group that allows to perform a release has been added. First, a staging build is pushed to `libs-staging-local`. Then, a promotion job can be invoked once the bits have been verified.

For the moment, only a local promotion is performed that triggers the automatic generation of the changelog.

### DIFF
--- a/ci/README.adoc
+++ b/ci/README.adoc
@@ -11,7 +11,7 @@ The pipeline can be deployed using the following command:
 
 [source]
 ----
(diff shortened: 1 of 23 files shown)


sdeleuze:
Amazing, thanks a lot. Let's sync today about the release process.

Issue #115, opened 2020-04-27
Add support for Log4j2


sdeleuze:
Related to #84.

sdeleuze:
See related https://github.com/oracle/graal/issues/808 closed issue.

sdeleuze:
Blocked on oracle/graal#2991.

sdeleuze:
Should be doable with GraalVM 21.1, see [this comment](https://github.com/oracle/graal/issues/2991#issuecomment-787828477) for more details.

sdeleuze:
Some hints are still missing so I reopen this issue.

sdeleuze:
Ok this one is more involved than expected since Log4j2 is really using a log of reflection. I have tried to [craft custom hints](https://github.com/sdeleuze/spring-native/commit/6dc5242434a41b6dad1d01f8848ea0aa3702a1fe) but even with that this is not super maintainable/complete and would require [native support on Log4j2 side](https://issues.apache.org/jira/browse/LOG4J2-2604) (preferred solution) or a dedicated `TypeProcessor` or `NativeConfiguration`. This would be doable but I think we have other priorities tight now, so I move this issue to the backlog, just keeping [`Log4jCoreHints`](https://github.com/spring-projects-experimental/spring-native/blob/main/spring-native-configuration/src/main/java/org/apache/logging/log4j/core/Log4jCoreHints.java) to avoid a hard breakage during compilation with some dependencies like Elasticsearch (see #398).

sdeleuze:
Superseded by https://issues.apache.org/jira/browse/LOG4J2-2649?focusedCommentId=17610488&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17610488.

PR #636, opened 2021-03-19
Add Lombok sample and Apache logging hints

joshlong:
adds support for Lombok's `@Log4j2` synthetic logs


sdeleuze:
Hi Josh, thanks for your contribution.

I think a sample will be required for that one, especially given the fact that Log4j2 has known issues on native and is currently not supported, see #115.

Please add a `commandlineruner-log4j2` sample with Log4j2 + Lombok to this PR in order to allow us to evaluate how far we are for getting that working.

joshlong:
Hi there, I've updated the PR with a `lombok-log4j2` sample. Thank you for your consideration.

spring-aot/src/main/java/org/springframework/nativex/type/TypeSystem.java
```diff
@@ -311,7 +323,8 @@ public Type resolve(org.objectweb.asm.Type type, boolean silent) {
 	public Type Lresolve(String desc, boolean silent) {
```
lyndseyBeil, line 323: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

sdeleuze:
@joshlong Thanks for the sample. I think we need to clarify what is the scope here, as pointed out in my previous comment, as far as I know Log4j2 is not supported in Spring Native (cf. #115) and the sample provided is using Logback underneath not Log4j2. I guess what happens in the provided use case is that `@Log4j2` generates code using Apache Commons Logging APIs provided by Spring. Is it your understanding as well?

sdeleuze:
Maybe `@Slf4j` would be a better fit here? Did you try it?

joshlong:
I'll try that out

joshlong:
yep, looks like switching out `@Slf4j` works too

sdeleuze:
I merged a reworked version, focusing the sample on Lombok but the hint on Apache Logging. Thanks for your contribution.

Issue #639, opened 2021-03-19
Custom Spring Data Elasticsearch repository fragment implementation not found when running in native


sothawo:
I am trying native with Spring Data Elasticsearch and am running into a problem when implementing a custom fragment for the Spring Data Elasticsearch repository.  It seems that the custom implementation class is not found (although added with `@TypeHint`).

A sample application showing this with a more detailed description can be found at https://github.com/sothawo/dataesnative

sdeleuze:
@christophstrobl Could you please have a look to the provided sample application to qualify it?

christophstrobl:
There seems to be an issue with picking up the component based on the naming of custom implementations. 
While we work on this issue, you can rename `PersonCustomRepositoryImp` to `PersonRepositoryImpl` which will allow it to be detected correctly. 
The provided sample app still won't work (due to missing configuration for one of the used libs), but the repository lookup won't be standing in your way any more.

sothawo:
thanks for the info; I'll get to the next errors as soon as I see them ;-) 
The workaround is alright in this example, but wehn there are multiple custom interfaces with their implementations, it would mean to merge the impls together into one file. 
But for my case - testing out Spring Data Elasticsearch with native - this is alright for the moment.

PR #661, opened 2021-03-24
Fix Spring Data custom repository fragment lookup.

christophstrobl:
This commit introduces a `Substitution` for the Spring Data `CustomRepositoryImplementationDetector` to directly use the `CandidateComponentsIndex` for repository fragment lookups and adds the missing type hint for the `RepositoryFragment` required to create the target repository.

This enables fragments following the pattern `FragmentInterfaceName` | `FragmentInterfaceNameImpl` to be discovered and registered correctly when annotated with `@Component` as outlined below.

```java
interface CustomRepository {
	// ... 
}

@Component
class CustomRepositoryImpl implements CustomRepository {
	// ... 
}

interface DomainTypeRepository extends CrudRepository<DomainType, String>, CustomRepository {
	// ... 
}
```

It also updates the data-elasticsearch example to cover the custom repository setup and retains functionality of the heritage configuration format (used in the data-mongodb).

On the long run it would be beneficial to revisit the `ClassPathScanningCandidateComponentProvider` for bean lookups using the `spring.components` index so it supports the in this PR substituted lookup.

----

Closes: #639

### DIFF
--- a/spring-native/src/main/java/org/springframework/nativex/substitutions/data/Target_CustomRepositoryImplementationDetector.java
+++ b/spring-native/src/main/java/org/springframework/nativex/substitutions/data/Target_CustomRepositoryImplementationDetector.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.nativex.substitutions.data;
+
+import java.io.IOException;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.oracle.svm.core.annotate.Substitute;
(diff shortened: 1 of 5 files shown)


spring-native/src/main/java/org/springframework/nativex/substitutions/data/Target_CustomRepositoryImplementationDetector.java
```diff
@@ -0,0 +1,90 @@
...
+
+
+@TargetClass(className = "org.springframework.data.repository.config.CustomRepositoryImplementationDetector", onlyWith = {OnlyIfPresent.class})
+public final class Target_CustomRepositoryImplementationDetector {
+
+	@Substitute
+	private Set<BeanDefinition> findCandidateBeanDefinitions(ImplementationDetectionConfiguration config) {
+
+		/* Using the index instead of ClassPathScanningCandidateComponentProvider with pattern.
+		 * Not sure why components are not found via the index as it should be configured on
```
sdeleuze, line 45: @aclement Could you please give us your advice on that question raised by @christophstrobl ?

aclement: I think that is worth an issue in its own right. I saw a similar thing whilst working with Josh Long - he was trying to use `ClassPathScanningCandidateComponentProvider` (IIRC) and it wasn't always able to use the components index - as if the combination of filters you set on it can cause it to not allow index usage.

sdeleuze: @christophstrobl Could you please create the issue suggested by Andy?

@aclement Do you think we should merge this PR as it is of should we wait the issue mention above to be fixed?

aclement: @sdeleuze I would say this improves things, let's integrate it. We have the issue for improving things further.

Issue #203, opened 2020-07-14
Add preliminary Spring Batch support


dsyer:
The Batch samples look like they are using generated JSON. It would be nice to tighten that up.

aclement:
Yes, I don't think it was pulled back into hints, I was just relieved when it was working. We should do that.

sdeleuze:
@aclement I am not sure we can properly support Batch in 0.8 yet, I have tried to polish a little bit the build config and I get this error message:
```
org.springframework.beans.factory.BeanDefinitionStoreException: @Configuration classes need to be marked as proxyBeanMethods=false. Found: [org.springframework.batch.core.configuration.annotation.ScopeConfiguration, org.springframework.batch.core.configuration.annotation.SimpleBatchConfiguration]
```

@mminella Could you please update Batch to use `proxyBeanMethods=false`?

mminella:
Depends on https://github.com/spring-projects-experimental/spring-native/issues/459

sdeleuze:
@mminella @aclement FYI batch is now using the build-time infrastructure like other samples, and to make it work I had to specify [this native hint](https://github.com/spring-projects-experimental/spring-native/blob/master/spring-native-samples/batch/src/main/java/com/example/batch/BatchApplication.java#L21).

Since 0.9.0 is getting closer and #459 does not look like a trivial change, I would like to propose to postpone both #459 and batch support to 0.9.1 is order to allow us to focus our time and energy on polishing bti on what is currently supported, and to mature properly batch support before announcing it. Any thoughts?

mminella:
@sdeleuze I have no objections as batch support is currently blocked on #459.

mminella:
@sdeleuze Why would that hint be added there and not to `BatchHints`?

sdeleuze:
@mminella It was temporary since I was not sure if it was Batch specific or not, but I can indeed move it to `BatchHints`.

After discussion with @aclement, we agreed that #459 and more generally batch support will require more work, so it is more reasonsable to move it to 0.9.1.

sdeleuze:
As discussed in https://github.com/spring-projects-experimental/spring-native/pull/667#discussion_r603449707, among other things, [the batch sample `build.sh`](https://github.com/spring-projects-experimental/spring-native/blob/master/samples/batch/build.sh) should be updated to avoid passing custom arguments in order to ensure the build is easily reproducible.

snicoll:
[...]

PR #667, opened 2021-03-25
Add Hints for the Spring Cloud Task project.

cppwfs:
It was tested using both ApplicationRunner and JobLauncherApplicationRunner(@EnableBatchProcessing).


samples/cloud-task/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+${PWD%/*samples/*}/scripts/compileWithMaven.sh -Dspring.native.build-time-properties-checks=default-include-all -Dspring.native.build-time-properties-match-if-missing=false -Dspring.native.factories.no-actuator-metrics=true && sleep 3 && ${PWD%/*samples/*}/scripts/test.sh
```
sdeleuze, line 3: Could you please share more about those custom properties? I am asking because samples are expected to be able to run with the default configuration in order to keep things simple for users.

sdeleuze (changes requested):

sdeleuze:
The `test.sh` script should not be needed since it is a [shared script](https://github.com/spring-projects-experimental/spring-native/blob/master/scripts/test.sh), you can probably remove it.

samples/cloud-task/build.sh
cppwfs: Updated the PR to remove the custom properties.   These were added to match those in the batch sample.    Do they need to be removed from the batch sample as well?

cppwfs:
Removed test.sh

samples/cloud-task/build.sh
sdeleuze: Thanks, for the batch sample this will be fixed as part of #203.

sdeleuze:
Polished and merged, thanks.

Issue #154, opened 2020-06-04
Document native developer experience (debug, etc.)


sdeleuze:
Inspired by spring-projects-experimental/spring-graalvm-native#125, we should leverage VS code [remote containers](https://code.visualstudio.com/docs/remote/containers) out of the box in samples, and also configure it to use [`-H:GenerateDebugInfo=1`](https://github.com/oracle/graal/blob/master/substratevm/DEBUGINFO.md) and `gdb` integration for debugging Spring Boot native images.

![image](https://user-images.githubusercontent.com/141109/84269115-0ee32f00-ab29-11ea-802b-d043934c76e3.png)

sdeleuze:
Spring Native is now superseded by Spring Boot 3 official native support, see [the related reference documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/native-image.html) for more details.

As a consequence, I am closing this issue, and recommend trying your use case with latest Spring Boot 3 version. If you still experience the issue reported here, please open an issue directly on the related Spring project (Spring Framework, Data, Security, Boot, Cloud, etc.) with a reproducer.

Thanks for your contribution on the experimental Spring Native project, we hope you will enjoy the official native support introduced by Spring Boot 3.

PR #674, opened 2021-03-26
Build ci images

snicoll:


snicoll:
Before we consider merging this, my proposal hasn't renamed some references to `springci/spring-native`. I was about to do that but found those `.devcontainer` directories in some samples. Can you explain what these are? Is that ok to reference to an hard-coded image there?

snicoll:
I've push forced a change where the JCMI url is different for the stable branch and the master branch (following https://github.com/spring-projects-experimental/spring-native/commit/9398859f9c585ec11e5e736f570e891cccc8e155)

snicoll:
This is a first step and I didn't want to modify the scripts themselves but concourse already checks out the remote repo automatically so there is no need to pass the `GRAALVM_BRANCH` argument. The script can be updated to just cd in the right directory.

sdeleuze:
`.devcontainer` directories are used on some samples to show how to use [Visual Studio Code Remote - Containers](https://code.visualstudio.com/docs/remote/containers). They have been introduced by @dsyer, I was tempted [to remove them](https://github.com/spring-projects-experimental/spring-native/commit/cbc49c0bc34a6a8b93ae866901150c01be8fbf0b) until we have not tackled #154 but Dave preferred to [re-introduced them](https://github.com/spring-projects-experimental/spring-native/pull/470). Maybe we can just reference the stable images here.

I saw you removed `docker/Dockerfile.spring-native-dev` but it is used by `run-dev-container.sh`. Can we can keep it and update `run-dev-container.sh` to specify `stable` or `dev` instead of specific versions for `--graalvm`?

snicoll:
> Maybe we can just reference the stable images here.

Images are tagged with the milestone to avoid mixing several images in several branches. Right now this is not very useful but will prove to be once we branch. I find it weird to reference a CI image like that outside of the concourse space and it is very likely to become out of sync. Perhaps I misunderstood the scope of the samples image?

> I saw you removed docker/Dockerfile.spring-native-dev but it is used by run-dev-container.sh

Ooops, I missed that. Yeah, I can have a look at that script again.

sdeleuze:
> Perhaps I misunderstood the scope of the samples image?

Conceptually I think those are not just images designed to be used on Concourse for samples, that's why the images have not "samples" in their name even if we install some samples related dependencies for this use case. Native imply more system dependencies so the usage of containers is more generalized than for a typical JVM arrangement. That's why `Dockerfile.spring-native-dev`, `run-dev-container.sh` and `.devcontainer` reference those images. We need to be able to work locally with a specific version of GraalVM to be able to debug and reproduce issues.

There is a wider discussion to have, maybe as part of #154, to shape our middle/long term strategy like if Buildpacks should provide dev oriented builders and we could maybe rationalize and focus on that. But that raises a lot of questions like IDE integration etc. so for now we need to keep this possibility and not make those CI images only designed to be used on the CI.

snicoll:
> that's why the images have not "samples" in their name even if we install some samples related dependencies for this use case.

Sounds like you stated something and the opposite in the same sentence. I get why the GraalVM image is not CI-only (or even Spring Native only) and they've been named accordingly but all of this is still primarily CI. Right now, even the `spring-native` images are under the `springci` org so I find this all very confusing. I think it's fine to reuse those images for another use case but making that a first-class (documented) concept is going to be a bit strange in the current form (current or in this PR).

> So for now we need to keep this possibility and not make those CI images only designed to be used on the CI.

With artifacts that are put in the image for the sole purpose of the samples, I don't see what you're expecting concretely.

sdeleuze:
> Sounds like you stated something and the opposite in the same sentence. I get why the GraalVM image is not CI-only (or even Spring Native only) and they've been named accordingly but all of this is still primarily CI. Right now, even the spring-native images are under the springci org so I find this all very confusing. I think it's fine to reuse those images for another use case but making that a first-class (documented) concept is going to be a bit strange in the current form (current or in this PR).

[This is documented](https://docs.spring.io/spring-native/docs/current/reference/htmlsingle/#how-to-contribute-using-container-build-env) and what we use for many months.

> but all of this is still primarily CI
It is not only for the CI, it is for both CI and Spring Native developers for usage on their local development platform.

As usual, sure there is room for improvement, but IMO that should be done as part of  #154, not this PR.

snicoll:
I've updated the proposal to use `spring-native` for the image used by the samples and the local dev. I've also updated `run-dev-container.sh` to use the new tagging scheme. 

As a result `milestone` is now copy/pasted in the pipeline and in the script. If we branch (or we switch to a new feature release), we'll have to remember to switch these as well. 

LMK what you think.

sdeleuze:
Looks great thanks, GraalVM `master` is still broken but this has nothing to do with your changes and GraalVM team has fixes currently processed in their pipeline but it could take a few days to make it to GraalVM `master`.

PR #690, opened 2021-03-31
Add initial set of configuration hints and sample for Spring Data REST

christophstrobl:
This PR adds the initial set of required hints to run a sample application using Spring Data REST and the HAL representation model.

```bash
http :8080/api/customers/1

{
    "_links": {
        "customer": {
            "href": "http://localhost:8080/api/customers/1{?projection}",
            "templated": true
        },
        "self": {
            "href": "http://localhost:8080/api/customers/1"
        }
    },
    "address": {
        "city": "Charlottesville",
        "state": "VA",
        "street": "4711 Some Place",
        "zipCode": "54321"
    },
    "firstname": "Dave",
    "lastname": "Matthews"
}
```
Along with Data REST specific hints the `allowWrite` attribute has been added to the `FieldHint` annotation (via ffdca18)

So far it is possible to expose the collection and entity endpoints. The `search` endpoint and others (like `profile`), as well as support for other media types and ALPS will be added subsequently moving forward.  

We'll potentially require additional changes in Data REST (eg. to fully support the `search` endpoint as the parameter name for the `@PathVariable` is not resolved from the method parameter name) and will upgrade to a more recent version once the upcoming Spring Data RC is out.

### DIFF
--- a/samples/data-rest/README.md
+++ b/samples/data-rest/README.md
@@ -1,0 +1,11 @@
+Spring Boot project with Spring Data REST.
+
+To build and run the native application packaged in a lightweight container:
+```
+mvn spring-boot:build-image
+docker-compose up
+```
+
+As an alternative, you can use `build.sh` (with a local GraalVM installation or combined with
+`run-dev-container.sh` at the root of `spring-native` project). See also the related issue
+[Take advantage of Paketo dev-oriented images](https://github.com/spring-projects-experimental/spring-native/issues/227).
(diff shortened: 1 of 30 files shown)


aclement:
The data-rest sample in there works for me. Thanks for implementing allowWrite I see you found my todo about supporting it when a use case came along :) I've pushed your commits (Hope that is OK??) and I further polished field descriptor handling, including implementing merging. I plan to do some more here to tidy things up further.

sdeleuze:
Since it has been merged I close this PR.

@christophstrobl Feel free to create other ones to refine or just push if that's straightforward.

PR #699, opened 2021-04-02
Add RabbitMQ Hints and Sample

garyrussell:
Simple RPC client and server.


snicoll:
@garyrussell we've got a user report in #702. I am hoping this PR addresses that use cases as well. If not we can reopen that issue.

garyrussell:
@snicoll Yes, this resolves that.

PR #726, opened 2021-04-13
support for the Kubernetes Java client so that we can build Kubernetes controllers and operators

joshlong:
This PR contributes support, documentation, and an example for the official Kubernetes Java client so that people can easily can build Kubernetes controllers and operators.


sdeleuze:
Thanks for this contribution Josh. It is for sure an important use case, but I am not sure Spring Native is the right place for such hints since we focus on the dependencies provided by start.spring.io and that's already a huge effort. We are working with GraalVM team on a repository for native configurations like that, my proposal is to wait for it to be available and contribute native configuration there. Another possibility is to create an issue/PR on the official Kubernetes Java client repository with JSON hints and maybe a feature if dynamic configuration is needed.

samples/petclinic-jpa/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<parent>
 		<groupId>org.springframework.experimental</groupId>
-		<artifactId>spring-native-sample-parent</artifactId>
+		<artifactId>sprigit stng-native-sample-parent</artifactId>
```
leccelecce, line 8: Pretty sure this is an accidental typing of a git command into your IDE window!

sdeleuze: Probably yes.

sdeleuze:
As discussed during today meeting, let's focus on providing a more flexible serialization support in order to make this kind of use case easier to support. @christophstrobl will likely create a related issue.

@joshlong We agreed on not integrating Kubernetes Java client hints in Spring Native so I will close this PR. Feel free to reach me to continue the discussion if needed.

christophstrobl:
The `TypeProcessor` is part of #733 adding a dedicated `ComponentProcessor` for JPA.

PR #733, opened 2021-04-13
Add jpa component processor.

christophstrobl:
This PR adds a `TypeProcessor` that traverses reachable structures from a given root type by inspecting the type itself, its annotations, fields and their annotations, constructors and their arguments & annotations, etc.

Whenever the processor discovers a `Type` (might as well be an annotation) it has not seen before it invokes a callback function.

For usage within `NativeConfiguration#computeHints` the `TypeProcessor` can build its own `NativeContext` that captures interactions to provide a `List` of `HintDeclaration` when done processing.

```java
public class DataRestHints implements NativeConfiguration {

	TypeProcessor restControllerProcessor = new TypeProcessor(
			(type, context) -> type.hasAnnotationInHierarchy(BASE_PATH_AWARE_CONTROLLER),
			(type, registrar) -> registrar.addReflectiveAccess(type, new AccessDescriptor(AccessBits.FULL_REFLECTION)),
			(annotation, registrar) -> {
				if (annotation.isPartOfDomain("org.springframework.web.bind.annotation")) {
					registrar.addReflectiveAccess(annotation, new AccessDescriptor(AccessBits.ANNOTATION));
				}
			}
	).named("RestControllerProcessor").excludeFieldInspection();
	
	@Override
	public List<HintDeclaration> computeHints(TypeSystem typeSystem) {
	
		return restControllerProcessor.use(typeSystem)
			.toProcessTypes(ts -> ts.findTypesAnnotated(BASE_PATH_AWARE_CONTROLLER, true)
				.stream()
				.map(ts::resolveName));
	}
}
```

When used within a `ComponentProcessor` implementation the processor directly writes the configuration to the given `NativeContext`.

```java
public class JpaComponentProcessor implements ComponentProcessor {

	private final TypeProcessor typeProcessor = new TypeProcessor(
			(type, context) -> {

				if (type.isPartOfDomain("sun.") || type.isPartOfDomain("jdk.")) {
					return false;
				}
				return true;
			},
			this::registerTypeInConfiguration,
			this::registerAnnotationInConfiguration
	).named("JpaComponentProcessor");

	@Override
	public void process(NativeContext imageContext, String componentType, List<String> classifiers) {

		Type domainType = imageContext.getTypeSystem().resolveName(componentType);
		typeProcessor.use(imageContext)
			.toProcessType(domainType);
	}

	// ...
}
```


spring-native-configuration/src/main/java/org/springframework/data/JpaComponentProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+
+					context.log(String.format("JpaComponentProcessor: detected final field %s for type %s. Setting allowWrite=true.", field.getName(), type.getDottedName()));
+					return new FieldDescriptor(field.getName(), true, true);
+				})
+				.collect(Collectors.toList());
+	}
+
+	interface JpaImplementation {
+
+		boolean isAvialable(NativeContext context);
```
aclement, line 117: spelling mistake

spring-native-configuration/src/test/java/org/springframework/nativex/util/NativeTestContext.java
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	@Override
+	public TypeSystem getTypeSystem() {
+		return typeSystem;
+	}
+
+	@Override
+	public void addReflectiveAccess(String key, Flag... flags) {
+
```
aclement, line 81: I don't like having AccessBits and having Flags, we need to rationalize that at some point. (But not necessarily in this PR).

spring-native-configuration/src/test/java/org/springframework/nativex/util/NativeTestContext.java
```diff
@@ -0,0 +1,204 @@
...
+		Integer value = 0;
+
+		// TODO: is this sufficient?
```
aclement, line 84: A bit like we have `AccessBits.getFlags(int)` we should centralize this kind of thing to avoid repeating it (`AccessBits.fromFlags(..)`?)

spring-native-configuration/src/test/java/org/springframework/nativex/util/TestTypeSystem.java
```diff
@@ -0,0 +1,92 @@
...
+ * Allows to {@link #excludePackages(String...)} based on their name - so make sure to {@link #clearExclusions()} during test setup.
+ *
+ * @author Christoph Strobl
+ */
+public class TestTypeSystem extends TypeSystem {
+
+	private Set<String> excludedPackages;
+
+	public TestTypeSystem() {
+		this(Collections.singletonList(new File("./target/classes").toString()));
```
aclement, line 41: Just checking, I had a problem the other day because I had "./target/classes" when it should have been "./target/test-classes" (or both depends on what I really wanted to resolve) - but I guess this is working for you.

spring-aot/src/main/java/org/springframework/nativex/type/Field.java
```diff
@@ -64,7 +65,18 @@ public Field(FieldNode node, TypeSystem typeSystem) {
 		}
 		return results == null ? Collections.emptyList() : results;
 	}
-	
+
+	/**
+	 * @return {@literal true} if considered a final field (ACC_FINAL || LDC).
+	 */
+	public boolean isFinal() {
+		return Opcodes.ACC_FINAL == node.access || Opcodes.LDC == node.access;
```
aclement, line 73: Is this quite right?  If I had a 'final static' field (`0x18`) this won't match? I also find it a little cryptic that you are using LDC as a shorthand for 'Opcodes.FINAL | Opcodes.PRIVATE' - I don't think I like that. It confuses the LDC opcode with the access flag values. Why isn't it written a bit more like the isSynthetic() one and testing for the final bit? `return (node.access & Opcodes.ACC_FINAL) != 0`.

spring-native-configuration/src/main/java/org/springframework/data/JpaComponentProcessor.java
christophstrobl: thanks will fix that

spring-native-configuration/src/test/java/org/springframework/nativex/util/NativeTestContext.java
christophstrobl: I'll move the code to `AccessBits` as I might need it in another place as well.

spring-aot/src/main/java/org/springframework/nativex/type/Field.java
christophstrobl: thanks - I'll change as you suggested.

spring-native-configuration/src/test/java/org/springframework/nativex/util/TestTypeSystem.java
christophstrobl: in this case it worked for me, but I get an error form `TypeSystem.scan` when the a project does not have tests. In that case `target/test-classes` not present and I run into the `IllegalStateException`. Happened when testing `samples/data-rest` with a `NativeConfiguration` that tries to lookup types with specific annotations via the `TypeSystem`.

spring-aot/src/main/java/org/springframework/nativex/type/TypeSystem.java
```diff
@@ -565,6 +565,12 @@ private void scanArchive(File f) {
 				// TODO resources?
 			}
 		} catch (IOException ioe) {
+			if(ioe instanceof NoSuchFileException) {
```
aclement, line 568: I put in a change to allow for empty classpath elements on a scan (the regular indexing was already coping with it) - so this should not be needed any more I think.

sdeleuze:
As discussed with @christophstrobl the `TypeProcessor` should probably live in `spring-aot` instead of `spring-native-configuration`.

@aclement Are you ok with the `TypeSystem` related changes?

sdeleuze:
Thanks for this great PR @christophstrobl !

Issue #469, opened 2021-02-07
Add Spring integration support


Crazyjavahacking:
I was wondering is Spring integration is supported or what is the current status of it?

sdeleuze:
No, it is not listed on https://github.com/spring-projects-experimental/spring-native/blob/master/spring-native-docs/src/main/asciidoc/support.adoc so it is not supported for now.

artembilan:
Here is the work to be done in Spring Integration project: https://github.com/spring-projects/spring-integration/issues/3502.

Once that one is committed and this project is ready to consume Spring Boot `2.5` as based line, I'll come back here with some possible additional work, e.g. fix that list in Docs that it is supported as first step 😉 .

Will add some sample, hints and everything we would need to confirm that Spring Integration is supported in Spring Native already.

PR #756, opened 2021-04-28
GH-469: Add Spring Integration support

artembilan:
Fixes https://github.com/spring-projects-experimental/spring-native/issues/469

* Introduce `org.springframework.integration.IntegrationHints`
for reflection and initialization types required in Spring Integration
for compatibility with Native
* Mention `spring-boot-starter-integration` as supported in the `support.adoc`
* Add `samples/integration` to demonstrate as much as possible of
Spring Integration in a simple app


samples/integration/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+	</parent>
+	<groupId>com.example</groupId>
+	<artifactId>integration</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+	<name>integration</name>
+	<description>GraalVM Spring Integration Demo</description>
+
+	<properties>
+		<java.version>11</java.version>
+		<spring-integration.version>5.5.0-SNAPSHOT</spring-integration.version>
```
artembilan, line 20: I need this version explicitly because there were some changes in SI already after an RC

spring-native-configuration/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationHints.java
```diff
@@ -1,27 +0,0 @@
...
-package org.springframework.boot.autoconfigure.integration;
-
-import org.springframework.nativex.type.NativeConfiguration;
-import org.springframework.nativex.hint.NativeHint;
-import org.springframework.nativex.hint.ResourceHint;
-
-@NativeHint(trigger = IntegrationAutoConfiguration.class,
-	resources = @ResourceHint(patterns="META-INF/spring.integration.default.properties")
-)
-public class IntegrationHints implements NativeConfiguration {
```
artembilan, line 26: Removed this class since there is no such a `META-INF/spring.integration.default.properties` resource in SI starting with version `5.5` any more

spring-native-configuration/src/main/java/org/springframework/integration/IntegrationHints.java
```diff
@@ -0,0 +1,123 @@
...
+								org.springframework.integration.graph.MessageGatewayNode.class,
+								org.springframework.integration.graph.MessageHandlerNode.class,
+								org.springframework.integration.graph.MessageProducerNode.class,
+								org.springframework.integration.graph.MessageSourceNode.class,
+								org.springframework.integration.graph.PollableChannelNode.class,
+								org.springframework.integration.graph.ReceiveCounters.class,
+								org.springframework.integration.graph.RoutingMessageHandlerNode.class,
+								org.springframework.integration.graph.SendTimers.class,
+								org.springframework.integration.graph.TimerStats.class,
+								org.springframework.integration.http.management.IntegrationGraphController.class,
```
artembilan, line 64: I assumed this one should be present in the reflection info according to `WebComponentProcessor`, but it is not visible somehow.
It has a `@RestController` on it and is registered as a bean functional way in the `IntegrationGraphControllerRegistrar`, so not sure what is going on here.

spring-native-configuration/src/main/java/org/springframework/integration/IntegrationHints.java
```diff
@@ -0,0 +1,123 @@
...
+		@TypeHint(
+				types = {
+						org.springframework.integration.core.GenericSelector.class,
+						org.springframework.integration.transformer.GenericTransformer.class,
+						org.springframework.integration.handler.GenericHandler.class,
+						java.util.function.Function.class,
+						java.util.function.Supplier.class,
+						kotlin.jvm.functions.Function0.class,
+						kotlin.jvm.functions.Function1.class,
+						kotlin.Unit.class
```
artembilan, line 101: I didn't try, but probably there is already some hint for some of these `java.util` and `kotlin.jvm`.
Happy to remove them if it works some other way!

samples/integration/pom.xml
sdeleuze: When next RC/release is planned?

spring-native-configuration/src/main/java/org/springframework/integration/IntegrationHints.java
sdeleuze: OK I will have a deeper look.

samples/integration/pom.xml
artembilan: May 18th, just before Spring Boot 2.5 GA: https://github.com/spring-projects/spring-integration/milestone/63 !

spring-native-configuration/src/main/java/org/springframework/integration/IntegrationHints.java
```diff
@@ -0,0 +1,123 @@
...
+				@ProxyHint(
+						types = {
+								org.springframework.integration.gateway.RequestReplyExchanger.class,
+								org.springframework.aop.SpringProxy.class,
+								org.springframework.aop.framework.Advised.class,
+								org.springframework.core.DecoratingProxy.class
+						}),
+				@ProxyHint(
+						types = {
+								org.springframework.integration.handler.AbstractReplyProducingMessageHandler.RequestHandler.class,
```
artembilan, line 84: @aclement ,

Apparently something had changes in the project because it worked for me yesterday after unwrapping strings into their normal classes, even for this nested one.

artembilan:
Forgot to mention, the `native-image-maven-plugin` fails for me on Windows. Just because it cannot find `cl.exe` in the PATH.
So, I had to install Microsoft Visual Studio: 
https://stackoverflow.com/questions/64197329/cl-exe-missing-when-building-native-app-using-graalvm
Then I tried to set PATH env var for respective `cl.exe`, but failed with many other missed stuff in the env:
https://stackoverflow.com/questions/84404/using-visual-studios-cl-from-a-normal-command-line

Perhaps GraalVM could run that recommended `vcvars64.bat` script from VC build before starting to build an image.
That's what VC people recommend instead of "polluting" ENV for many VC vars: https://social.msdn.microsoft.com/Forums/vstudio/en-US/2ebc59de-dc8e-426f-b1ca-885c2709df5f/visual-c-installed-but-cannot-find-clexe?forum=vcgeneral

Right now I give up in favor of working variant with the `spring-boot:build-image`.

sdeleuze:
In https://docs.spring.io/spring-native/docs/current/reference/htmlsingle/#_build_the_native_application_2 we document: On Windows make sure to use x64 Native Tools Command Prompt as recommended in the [GraalVM native-image prerequisites](https://www.graalvm.org/reference-manual/native-image/#prerequisites). Is that the way you run your build?

artembilan:
Thanks @sdeleuze !

That worked for me: when I run `vcvars64.bat` from command line and `cd`ed to my sample I definitely could build it! Now I see that cool `integration.exe` in the `\target`! 😸

sdeleuze:
@artembilan Glad to hear that. Should I merge this PR for inclusion in 0.10.0 from your POV?

artembilan:
We can merge now or wait until Boot 2.5 GA in a couple weeks.
Plus I see CI has failed for this PR. I'm not familiar with Concourse and complex Maven build, so would be great if someone take a look over there and let me know what is wrong and what to fix if that.

Thanks

Issue #854, opened 2021-06-22
Proxy entries for annotations on interface not created


phejl:
When controller extends interface which defines the contract (ie openapi generator generates such) the annotations present on the interface are not processed and thus proxy entries containing `SynthesizedAnnotation` are not created.

phejl:
I'll try to provide a patch.

phejl:
Hopefully fixed in https://github.com/spring-projects-experimental/spring-native/pull/857

aclement:
PR is merged

PR #857, opened 2021-06-23
#854 - Proxy entries for annotations on interface not created

phejl:


aclement (approved):
It looks ok. unit tests are fine but I do find tweaking an existing sample so that it actually needs your code to work properly is a good way to get a positive feeling that it'll behave in a native-image.

aclement:
I merged it.

sdeleuze:
@phejl Thanks for your contributions and feel free to create PRs without related issue, we can set PR milestone and they will be visible in the changelog.

phejl:
@sdeleuze Thank you. Noted.

Issue #861, opened 2021-06-24
Proxies for @Validated components not prepared


phejl:
When component (such as rest controller) is annotated with `@Validated` the app fail at runtime as the proxy is not prepared. Same happens when `@Validated` is used on interface component implements.

phejl:
Suggested fix https://github.com/spring-projects-experimental/spring-native/pull/866

PR #866, opened 2021-06-28
#861 - Proxies for @Validated components not prepared

phejl:


spring-native-configuration/src/main/java/org/springframework/ValidatedComponentProcessor.java
```diff
@@ -0,0 +1,65 @@
...
+	@Override
+	public boolean handle(NativeContext imageContext, String componentType, List<String> classifiers) {
+		Type type = imageContext.getTypeSystem().resolveName(componentType);
+		return type.isAtValidated(true);
+	}
+
+	@Override
+	public void process(NativeContext imageContext, String componentType, List<String> classifiers) {
+		Type type = imageContext.getTypeSystem().resolveName(componentType);
+		List<String> transactionalInterfaces = new ArrayList<>();
```
aclement, line 44: the name of this looks like a leftover from the transactional component processor it was a copy of?

spring-native-configuration/src/main/java/org/springframework/ValidatedComponentProcessor.java
```diff
@@ -0,0 +1,65 @@
...
+				hasInterfaceMethods = true;
+			}
+		}
+		if (!transactionalInterfaces.isEmpty() && hasInterfaceMethods) {
+			transactionalInterfaces.add("org.springframework.aop.SpringProxy");
+			transactionalInterfaces.add("org.springframework.aop.framework.Advised");
+			transactionalInterfaces.add("org.springframework.core.DecoratingProxy");
+			imageContext.addProxy(transactionalInterfaces);
+			imageContext.log(ValidatedComponentProcessor.class.getSimpleName() + ": creating proxy for these interfaces: " + transactionalInterfaces);
+		} else if (!type.isInterface()) {
```
aclement, line 58: We should push this check up - there is no need to gather and walk through the interfaces if the type cannot be JDKProxied. (And this should be fixed in the transactional comp processor too).

spring-native-configuration/src/test/java/org/springframework/validated/ValidatedComponentProcessorTest.java
```diff
@@ -0,0 +1,123 @@
+package org.springframework.validated;
```
aclement, line 1: I think it is perhaps also worth tweaking the existing validator sample we have with a bit of code that needs this component processor (maybe its JDKproxy and AOTproxy variants, but particularly the latter) to verify this all works in a native image.

aclement (commented):
Thanks for taking the time to contribute - I mentioned a few things we could do to make it even better if you want to?

phejl:
> Thanks for taking the time to contribute - I mentioned a few things we could do to make it even better if you want to?

Yeah, sure. I'm not a real expert in Spring internals. Just give me some time to check your comments. Thanks!

spring-aot/src/main/java/org/springframework/nativex/type/Type.java
```diff
@@ -242,6 +243,16 @@ public String toString() {
 		return interfaces;
 	}
 
+	public Type[] getAllInterfaces() {
```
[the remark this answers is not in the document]
phejl: Should be fixed. I've also added `getImplementedInterfaces` to get only direct interfaces of the class (for proxying).

spring-native-configuration/src/main/java/org/springframework/ValidatedComponentProcessor.java
phejl: Yes. Thanks for catching that. Fixed.

spring-native-configuration/src/main/java/org/springframework/ValidatedComponentProcessor.java
phejl: I'm not sure about this as we first check for interfaces (and implemented methods) and only otherwise (else) if it is a class we create aot proxy. Am I missing something?

spring-native-configuration/src/test/java/org/springframework/validated/ValidatedComponentProcessorTest.java
phejl: I'll try that tomorrow.

spring-native-configuration/src/main/java/org/springframework/ValidatedComponentProcessor.java
aclement: Before we had the ability to proxy classes we just 'had a go' at creating a JDKProxy for the interfaces a class implemented. This wasn't really correct but we got away with it sometimes. We added class proxies support and shoved them in as an afterthought, at least on the transactional component processor (in that processor there is still digging to do there as different samples seem to be satisfied with one vs the other - see how it isn't `else if` in the transactional processor, it will sometimes create both!).   I think this is really why I wanted to see some changes to the real sample around validation - add a few variant types with validation expressed in different ways (on the class itself, on interfaces, etc) - that's how we can truly discover if the component processor is doing the job it needs too.

aclement: (If we aren't careful we can proxy too much - and it may work but images will be unnecessarily bloated with stuff that wasn't required)

phejl: I agree. Once this is done I would like to check the transactional case and whether similar rules apply.

spring-native-configuration/src/test/java/org/springframework/validated/ValidatedComponentProcessorTest.java
phejl: I've added classes to validator sample. Now I observe the strange thing where I can see AOT proxy classes being generated yet the resulting image complain they are not available :thinking:

sdeleuze:
@aclement I will let you merge this one if you think that's ready, or request additional changes.

phejl:
Thanks for merging @aclement. I'm still suffering strange errors in sample project whcih I'm tryingto deal with.
Although the proxy is generated:
`[INFO] Generating class file bytes for a proxy named com.example.validator.hierarchy.ValidatedComponent$$SpringProxy$422694c5`
It is not found (due to CNFE) at native image runtime:
`java.lang.ClassNotFoundException: com.example.validator.hierarchy.ValidatedComponent$$SpringProxy$422694c5`
`No suitable proxy found with name com.example.validator.hierarchy.ValidatedComponent$$SpringProxy$422694c5`

phejl:
Aha the cause is
```
[INFO]     [creator]     WARNING: Could not resolve com.example.validator.hierarchy.ValidatedComponent$$SpringProxy$422694c5 for reflection configuration. Reason: java.lang.UnsupportedClassVersionError: com/example/validator/hierarchy/ValidatedComponent$$SpringProxy$422694c5 has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0.
```

PR #891, opened 2021-07-06
Spring Cloud Square

joshlong:
This PR adds support for building Spring Cloud Square-powered OkHttp and Retrofit-based HTTP clients, optionally also supporting client-side loadbalancing.


samples/square/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    <description>Demo project for Spring Boot</description>-->
+
+    <parent>
+        <groupId>org.springframework.experimental</groupId>
+        <artifactId>spring-native-sample-parent</artifactId>
+        <version>0.10.2-SNAPSHOT</version>
+        <relativePath>../maven-parent/pom.xml</relativePath>
+    </parent>
+    <groupId>com.example</groupId>
+    <artifactId>async</artifactId>
```
sdeleuze, line 26: Could we change that to be = to the directory name "square"?

samples/square/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    <version>0.0.1-SNAPSHOT</version>
+    <properties>
+        <java.version>1.8</java.version>
+        <spring-cloud.version>2020.0.3</spring-cloud.version>
+
+        <!--
+        TODO switch this to M2 when its released.
+        -->
+        <spring-cloud-square.version>
+            0.4.0-SNAPSHOT
```
sdeleuze, line 36: Let's wait `0.4.0-M2` to merge this PR.

sdeleuze (changes requested):
Thanks for this contribution, I added a few comments + could you please check why the PR CI build is broken?

ci/pipeline.yml
```diff
@@ -128,11 +128,6 @@ resources:
       repository: spring-native
```
aclement, line 128: why does the PR modify our ci pipeline?

spring-native-configuration/src/main/java/org/springframework/cloud/square/SquareHints.java
```diff
@@ -0,0 +1,47 @@
...
+	*
+	* @author Josh Long
+	* @author Andy Clement
+	*/
+@NativeHint(
+	trigger = RetrofitAutoConfiguration.class,
+	options = "-H:+AddAllCharsets --enable-url-protocols=http,https",
+	types = {
+		@TypeHint(typeNames = {
+			"jdk.vm.ci.meta.JavaKind$FormatWithToString[]",
```
aclement, line 26: @sdeleuze I recall this one was weird but I thought it was fixed by something in GraalVM wasn't it?

spring-native-configuration/src/main/java/org/springframework/cloud/square/SquareHints.java
```diff
@@ -0,0 +1,47 @@
...
+			"java.lang.reflect.AnnotatedElement[]",
+			"java.lang.reflect.GenericDeclaration[]",
+			"com.oracle.svm.core.hub.Target_java_lang_constant_Constable[]",
```
aclement, line 29: These 3 are duplicates are what you have in the square sample project reflect-config.json - feels weird to have them in both places

spring-native-configuration/src/main/java/org/springframework/cloud/square/SquareHints.java
sdeleuze: I am not sure, we still have it in `LoadBalancerHints`, likely to be tested to check.

ci/pipeline.yml
joshlong: no earthly idea. please feel free to ignore any changes to the CI pipeline

aclement:
I've polished this up a bit in https://github.com/spring-projects-experimental/spring-native/pull/945 - but that won't go in until M2 of square is out (because it doesn't work for M1 and I'd rather not have a snapshot dependency there), we may hold it off until 0.11 as well because that changes many things.

PR #895, opened 2021-07-06
Have better out of the box support for logback

phejl:


sdeleuze:
@phejl Hi, could you please provide more details about the new capabilities provided by this PR?

Based on [Spring Boot documentation](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/logging.adoc):
> In addition, a legacy base.xml file is provided for compatibility with earlier versions of Spring Boot.

So I would tend to think `base.xml` is not required to be included, for `defaults.xml` I am not sure but I think it is now done programmatically.

@wilkinsona Could you please confirm that `base.xml` is not required for a typical Spring Boot 2.5 app? What about `defaults.xml`, based on the documentation is seems needed but maybe such configuration is now performed programmatically?

phejl:
> @phejl Hi, could you please provide more details about the new capabilities provided by this PR?
> 
> Based on [Spring Boot documentation](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/logging.adoc):
> 
> > In addition, a legacy base.xml file is provided for compatibility with earlier versions of Spring Boot.
> 
> So I would tend to think `base.xml` is not required to be included, for `defaults.xml` I am not sure but I think it is now done programmatically.

@sdeleuze Aha, that might be my misuderstanding. As most of the online resources still uses incluson of `defaults.xml` (including my app). Hence profiler suggested it should be included. Typehints are suggested for common things one ususally customize (console appender, rolling files).

Adds support for simple condisions in form `isDefined('VAR')` and `isNull('VAR')` as I believe this might cover major portion of use cases of Janino which would otherwise fail in native image as it is creating bytecode at runtime and loading it as class.

sdeleuze:
Interesting, so does this means we can now support basic `logback.xml` files by skipping Janino for such use cases?

In the documentation update you mentioned:
> Logback XML configuration via `-Dlogging.config` or `--logging.config` is supported but you need to enable XML support

I recently removed the conditional activation of the Logback substitution based on the XML flag since we were not supporting it, if we do we should maybe bring back `RemoveXmlSupport.class` [here](https://github.com/spring-projects-experimental/spring-native/commit/d7c785d00677e9d7bc5e41ebfb015bd253162a6e#diff-1d64e49dffd4742642e0b6a63325f14735f6102967b36bea5d2409eca6f31fb0R37).

phejl:
> Interesting, so does this means we can now support basic `logback.xml` files by skipping Janino for such use cases?

I thought the problem with `logback.xml` on classpath was it was triggering some unwanted build time initializations. My use case I have tested to be working was like `./my-native-image --logging.config=/my/logback-config.xml` (with enabled xml support) containing simple conditions.

> 
> In the documentation update you mentioned:
> 
> > Logback XML configuration via `-Dlogging.config` or `--logging.config` is supported but you need to enable XML support
> 
> I recently removed the conditional activation of the Logback substitution based on the XML flag since we were not supporting it, if we do we should maybe bring back `RemoveXmlSupport.class` [here](https://github.com/spring-projects-experimental/spring-native/commit/d7c785d00677e9d7bc5e41ebfb015bd253162a6e#diff-1d64e49dffd4742642e0b6a63325f14735f6102967b36bea5d2409eca6f31fb0R37).

Hm, in my case of log file argument this code is actually bypassed (`org.springframework.boot.logging.logback.LogbackLoggingSystem:185`) as long as XML is enabled.

phejl:
@sdeleuze I looked again at the problem with `logback.xml` bundled as resource (not something this PR intend to solve, but anyway). If I understand things correctly the problematic part is `org.slf4j.impl.StaticLoggerBinder`'s static initializer calling `SINGLETON.init();`. I'm afraid we can't really push all this to build time. It would initialize custom things such as `ch.qos.logback.core.spi.PropertyDefiner`s in build time when it might not be possible to really evaluate the value (might be deployment specific). On the other hand I assume we don't want to do everything at runtime. I guess the solution would be to postpone that `init()` method to be invoked lazily. I'm not sure how to do that without touching it's source code as it seems the `@Substitution` is not applied at the time class initializers are invoked at build time.

wilkinsona:
> @wilkinsona Could you please confirm that base.xml is not required for a typical Spring Boot 2.5 app? What about defaults.xml, based on the documentation is seems needed but maybe such configuration is now performed programmatically?

A Boot app that only uses properties to configure its logging won't use `base.xml` or `defaults.xml`. If someone is using XML to configure Logback then they may import `defaults.xml` but they do not have to do so. They may also choose to import `console-appender.xml` or `file-appender.xml`. No one should be using `base.xml` any more, particularly not if they're on the leading edge and using Spring Native.

phejl:
> > @wilkinsona Could you please confirm that base.xml is not required for a typical Spring Boot 2.5 app? What about defaults.xml, based on the documentation is seems needed but maybe such configuration is now performed programmatically?
> 
> A Boot app that only uses properties to configure its logging won't use `base.xml` or `defaults.xml`. If someone is using XML to configure Logback then they may import `defaults.xml` but they do not have to do so. They may also choose to import `console-appender.xml` or `file-appender.xml`. No one should be using `base.xml` any more, particularly if they're on the leading edge and using Spring Native.

@wilkinsona Thanks for your input! Should the patch be updated to allow access to `defaults.xml`, `console-appender.xml` and `file-appender.xml` (ommitting `base.xml`) as a result?

wilkinsona:
I think that's a question of what scope the Spring Native team want their Logback support to have, so I'll defer to them.

sdeleuze:
@phejl I see in the updated documentation:
> Logback XML configuration via `-Dlogging.config` or `--logging.config` is supported but you need to enable <<spring-aot-configuration, XML support>>.

What prevents us to support `logback.xml` provided in the classpath if it uses the simple constructs supported in the PR?

sdeleuze:
@phejl Sorry I missed you comment on https://github.com/spring-projects-experimental/spring-native/pull/895#issuecomment-876198704 where you already explained that.

Please add a sample in order to allow me to merge this PR.

phejl:
@sdeleuze 
> Please add a sample in order to allow me to merge this PR.

Sure I'll try. I just need some more spare time :)

phejl:
@sdeleuze 
> Please add a sample in order to allow me to merge this PR.

Sample added.

samples/logger/pom.xml
```diff
@@ -32,6 +32,11 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter</artifactId>
 		</dependency>
+		<dependency>
```
sdeleuze, line 35: Could you please share more context about why the janino dependency is needed?

phejl: To trigger (enable) conditional processing and prove that the log config with conditions (conditions in limited form of `isDefined` and `isNull`) works ok with native image. Note the log config contains such a condition.

phejl:
@sdeleuze Please, is there anything more needed from my side?

spring-native/src/main/java/org/springframework/nativex/substitutions/JaninoIsAround.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class JaninoIsAround implements BooleanSupplier {
+
+	@Override
+	public boolean getAsBoolean() {
+		try {
+			return Class.forName("org.codehaus.janino.ScriptEvaluator", false, getClass().getClassLoader()) != null;
+		} catch (ClassNotFoundException | NoClassDefFoundError ex) {
+			return false;
+		}
```
ttddyy, line 29: This could be simplified to 

```java
return ClassUtils.isPresent("org.codehaus.janino.ScriptEvaluator", getClass().getClassLoader());
```

phejl: @ttddyy Thanks for the hint. Applied.

phejl:
@sdeleuze I agree. The only thing I was doing from time to time was rebasing to recent main. I was even running my app with the changes (and load tested it) and I havn't found any issue regarding logging.

Issue #896, opened 2021-07-07
Add support for configurable main class in aot plugins 


dsyer:
```
[ERROR] Failed to execute goal org.springframework.experimental:spring-aot-maven-plugin:0.10.1:generate (generate) on project micro: Build failed during Spring AOT code generation: Unable to find a single main class from the following candidates [com.example.reactor.ReactorApplication, com.example.mini.MiniApplication, com.example.micro.MicroApplication, com.example.manual.ManualApplication, com.example.func.FuncApplication, com.example.func.CuncApplication, com.example.func.BuncApplication, com.example.empt.EmptyApplication, com.example.demo.DemoApplication, com.example.boot.BootApplication] -> [Help 1]
```

ttddyy:
FYI: related stacktrace

```
...
Caused by: java.lang.IllegalStateException: Unable to find a single main class from the following candidates [com.example.nativer2dbcproxy.NativeR2dbcProxyApplication, com.example.nativer2dbcproxy.AnotherApplication]
    at org.springframework.boot.loader.tools.MainClassFinder$SingleMainClassCallback.getMainClassName (MainClassFinder.java:432)
    at org.springframework.boot.loader.tools.MainClassFinder$SingleMainClassCallback.access$100 (MainClassFinder.java:403)
    at org.springframework.boot.loader.tools.MainClassFinder.findSingleMainClass (MainClassFinder.java:110)
    at org.springframework.boot.loader.tools.MainClassFinder.findSingleMainClass (MainClassFinder.java:94)
    at org.springframework.aot.nativex.ConfigurationContributor.getMainClass (ConfigurationContributor.java:114)
    at org.springframework.aot.nativex.ConfigurationContributor.contribute (ConfigurationContributor.java:78)
    at org.springframework.aot.BootstrapCodeGenerator.generate (BootstrapCodeGenerator.java:75)
    at org.springframework.aot.maven.GenerateMojo.execute (GenerateMojo.java:65)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:210)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
[...]

PR #900, opened 2021-07-10
Add main class name to BuildContext

ttddyy:
Add the main class name to `BuildContext` to pass it down to the code generation.

This change provides a way to support a case when multiple main methods exist in the classpath. (#896)
Once the gradle and maven plugins are updated to take a main class name, `BuildContext` can hold the value and the code generation step can reference the specified main class.


ttddyy:
Closing in favor of https://github.com/spring-projects-experimental/spring-native/pull/907

PR #921, opened 2021-07-22
Add more support with Spring for Apache Kafka

artembilan:
* Add missed `JsonDeserializer` into type hints
* Add `@InitializationHint` for some `spring-kafka` JSON components
which need to initialize their `static` props for `ClassUtils.isPresent()`
* Improve `kafka` sample for JSON data interaction
* Add a `@TypeHint` for end-user data classes used in the sample
* Improve `docker-compose.yml` for extra image of the current `kafka` sample

### DIFF
--- a/samples/kafka/src/main/java/com/example/kafka/KafkaApplication.java
+++ b/samples/kafka/src/main/java/com/example/kafka/KafkaApplication.java
@@ -10,7 +10,12 @@ import org.springframework.kafka.annotation.KafkaListener;
 import org.springframework.kafka.config.TopicBuilder;
 import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.nativex.hint.TypeHint;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@TypeHint(types = KafkaApplication.Greeting.class)
 @SpringBootApplication
 public class KafkaApplication {
 
@@ -19,7 +24,7 @@ public class KafkaApplication {
 	}
 
 	@KafkaListener(id = "graal", topics = "graal")
-	public void listen(String in) {
+	public void listen(Greeting in) {
 		System.out.println("++++++Received:" + in);
 	}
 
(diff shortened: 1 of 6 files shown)


artembilan:
Depends on: https://github.com/spring-projects/spring-kafka/pull/1881

samples/kafka/src/main/java/com/example/kafka/KafkaApplication.java
```diff
@@ -29,15 +34,38 @@ public NewTopic topic() {
...
+			ConsumerFactory<String, Greeting> cf) {
+
+		cf.addListener(new ConsumerFactory.Listener() {
 
-		cf.addListener(new ConsumerFactory.Listener() { });
+		});
 		return args -> {
-			template.send("graal", "foo");
+			template.send("graal", new Greeting("Hello from GraalVM!"));
 			System.out.println("++++++Sent:foo");
```
garyrussell, line 45: Change this to match?

garyrussell (changes requested):

garyrussell:
Didn't you test it?

You need to update the verify.sh https://github.com/spring-projects-experimental/spring-native/blob/main/samples/kafka/verify.sh

artembilan:
Huh? What is that? Never used those `.sh`. Tried `build.sh` once, but it was silent and in the end it produces an `.exe` which is not always working on Windows.
Yeah.. Thanks for the pointer! Let me see what that `verify.sh` does for us and fix it respectively! 
Docs don't mention this file though...

garyrussell:
See https://github.com/spring-projects-experimental/spring-native/blob/main/scripts/test.sh

called from https://github.com/spring-projects-experimental/spring-native/blob/main/samples/kafka/build.sh

artembilan:
Repushed after rebasing to `main`.
Checked locally with those `build.sh` and `verify.sh`: works well even on Windows for me ! 😄

garyrussell:
Hmmm - on MacOS:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Native reflection configuration for org.springframework.kafka.support.serializer.JsonDeserializer is missing.
```

artembilan:
Correct. You must build and install locally the main project first.
That `JsonDeserializer` was missed in the original `KafkaHints`, so I added it as a part of my PR.

garyrussell:
Ah; ok; sorry.

garyrussell (approved):
LGTM; however since it needs a spring-kafka snapshot, it can't be merged unless there are no plans for a release before August 16 (when spring-kafka 2.7.5 will be released). But I could consider releasing it earlier than that, if needed.

artembilan:
Well, there is no strong dependency on spring-kafka in the main project. Right now it is only a matter of a sample. `spring-native` can be released as is. 
We may find something else though for better compatibility before that August 16th...

PR #932, opened 2021-07-27
Use target class version for proxy

phejl:
Uses the classfile version of the proxy to avoid confusion on different JDK in CLI and build image.


sdeleuze:
@aclement See https://github.com/spring-projects-experimental/spring-native/pull/866#issuecomment-876368234.

PR #933, opened 2021-07-27
Improve transactional proxy detection to generate needed proxy hints

phejl:
Improved transactional proxy detection for super classes and super interfaces so proxy hints are properly and automatically generated in these cases. This covers cases where `@Transactional` annotation is placed on interface implemented or its super interfaces or where the annotation is placed on super class.

### DIFF
--- a/samples/data-jpa/src/main/java/app/main/hierarchy/ComponentWithTransactionalInterface.java
+++ b/samples/data-jpa/src/main/java/app/main/hierarchy/ComponentWithTransactionalInterface.java
@@ -1,0 +1,14 @@
+package app.main.hierarchy;
+
+import org.springframework.stereotype.Component;
+
+// needs jdk proxy
+@Component
+public class ComponentWithTransactionalInterface implements TransactionalInterface {
+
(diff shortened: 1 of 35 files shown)


aclement:
I wouldn't mind seeing some text, ideally an issue, that describes the actual problem that this code fixes (I'm sure there is a problem, but it isn't clear).  When browsing a release I think people see issues resolved, not PRs merged.

sdeleuze:
People see the PRs if we set the milestone on it, and I prefer to not have duplicated issue/PR so if @phejl You can update this PR description or add a more detailed comment here I think that would be fine.

phejl:
> @phejl you can update this PR description or add a more detailed comment here, I think that would be fine.

@sdeleuze Yes I tried. Hopefully it is better explained now.

sdeleuze:
@aclement Could you please review this PR?

aclement:
Finally merged! I had to adjust the changes from the component processor into the new native configuration processor. I adjusted the test cases and merged those too. Thanks for the contribution!

Issue #869, opened 2021-06-29
Support for @EntityListeners


goafabric:
Now with version 0.10.1 adding Transactional out of the box for classes .. thats a very good step in the right direction.
What is currently not working is the Annoation of @EntityListners
e.g 

@EntityListeners(AuditJpaListener.class)
public class PersonBO{}

The class just never gets called .. funny enough no error during aot compilation or bootstrap .. just seems to be ignored.
Works of course in JVM mode.

Would be nice to get this fixed, or a hint into the right direction.
Thx

phejl:
@goafabric I'm not an expert but only thing I needed to do in my simple test case was to add TypeHint for the listener class. Sure it would be nicer if this was detected automatically. Just providing workaround and checking with you whether it is the same problem you observe.
Try to add `@NativeHint(types = @TypeHint(types = AuditJpaListener.class))` to your application class.

phejl:
I'll try to provide a fix.

goafabric:
Hi @phejl,
thanks for your support .. yes that might actually work .. in Quarkus you have to do something similar.
I just tried to avoid Typehints in general ,.. but i see that you also provided a fix ...

goafabric:
@phejl 
do you also have an Idea for the not working ApplicationContextAware

phejl:
> @phejl
> do you also have an Idea for the not working ApplicationContextAware

@goafabric Hm, I don't think it is directly related. Would you please file a separate issue with details of your usecase?

goafabric:
> > @phejl
> > do you also have an Idea for the not working ApplicationContextAware
> 
> @goafabric Hm, I don't think it is directly related. Would you please file a separate issue with details of your usecase?

Its indirectly related .. i opened the #949 for this

sdeleuze:
Let see how that goes with 0.11.

venkataklagisetty:
hi ,I got same problem with entityListener, is that resolved?

snicoll:
@venkataklagisetty this issue is open so it's likely not resolved.

venkataklagisetty:
> @venkataklagisetty this issue is open so it's likely not resolved.

@snicoll  Thanks for confirming that, is there any alternate approach available to resolve this issue now?

PR #934, opened 2021-07-27
#869 - Support for @EntityListeners 

phejl:


sdeleuze:
@phejl Thanks for this PR, does it take in account/solve the point raised in https://github.com/spring-projects-experimental/spring-native/issues/869#issuecomment-879664759 ?

spring-native-configuration/src/main/java/org/springframework/data/JpaComponentProcessor.java
```diff
@@ -70,6 +72,11 @@ public void process(NativeContext imageContext, String componentType, List<Strin
 
 		Type domainType = imageContext.getTypeSystem().resolveName(componentType);
 		typeProcessor.use(imageContext).toProcessType(domainType);
+
+		for (String listener : domainType.findAnnotationValue(ENTITY_LISTENERS, false, false)) {
+			Type listenerType = imageContext.getTypeSystem().Lresolve(listener);
+			imageContext.addReflectiveAccess(listenerType.getDottedName(), AccessBits.FULL_REFLECTION);
```
christophstrobl, line 78: Does it make sense to be more specific here in terms of methods that require access? 
Maybe something like:
```java
List<MethodDescriptor> methodDescriptors = listenerType
	.getMethods(method -> method.hasAnnotations()) // or even just javax ones
	.stream()
	.map(it -> new MethodDescriptor(it.getName(), Arrays.asList(it.asConfigurationArray())))
	.collect(Collectors.toList());

imageContext.addReflectiveAccess(listenerType.getDottedName(), new AccessDescriptor(AccessBits.LOAD_AND_CONSTRUCT, methodDescriptors, Collections.emptyList()));
```			
Otherwise I think it's fine for now to stick with `FULL_REFLECTION`.

phejl: > @phejl Thanks for this PR, does it take in account/solve the point raised in [#869 (comment)](https://github.com/spring-projects-experimental/spring-native/issues/869#issuecomment-879664759) ?

I asked for a separate issue.

phejl: > Does it make sense to be more specific here in terms of methods that require access?

Thanks! Let me check that.

phejl: @christophstrobl I've updated the PR. Thanks for your suggestion!

sdeleuze:
Merged via 5a8d9d423248822052db170b1aa0b7dd1259a8cd thanks.

PR #946, opened 2021-07-31
Fix Gradle task name in sample README

darxriggs:
It was accidentally using Maven syntax.

@pivotal-cla This is an Obvious Fix


sdeleuze:
Thanks

Issue #968, opened 2021-08-12
Maven code generation process does not add bytebuddy to the classpath


snicoll:
When generating the bootstrap code using the maven plugin, bytebuddy isn't added to the classpath of the forked process. Adding the dependency to the project itself fixes it.

This can be reproduced by trying to build the `jdbc-tx` sample (once it has moved to `SpringAotApplication`.

/cc @aclement

ttddyy:
I created this PR(#971) which fixes this issue.

After applying this change, the `jdbc-tx` sample still fails.

For quick fix, I copied these hints from `org.springframework.boot.autoconfigure.jdbc.JdbcHints`. Then, the sample works with `0.11.x` branch.

```java
@NativeHint(types = {
		@TypeHint(types= DatabaseMetaData.class, methods= @MethodHint(name="getDatabaseProductName")),
		@TypeHint(types= {ConcurrentBag.IConcurrentBagEntry[].class, ConcurrentBag.IConcurrentBagEntry.class, Statement.class, Statement[].class}),
		@TypeHint(types = {HikariDataSource.class}),
		@TypeHint(types = HikariConfig.class, typeNames = "com.zaxxer.hikari.HikariConfigMXBean", access = AccessBits.FULL_REFLECTION)})
@NativeHint(trigger= DataSourceAutoConfiguration.class, resources = {
		@ResourceHint(patterns = {"schema.sql", "data.sql" })
})
```

snicoll:
Looks like the hint are not processed in the branch anymore, that is interesting. Thanks @ttddyy!

Closing in favor of PR #971

PR #971, opened 2021-08-13
Add ByteBuddy to the classpath of maven code generation process

ttddyy:
In `0.11.x` branch, when `GenerateMojo` invokes `BootstrapCodeGeneratorRunner`, it failed to find bytebuddy. (#968)

This is because when the `ProxyGenerator`, called by `ConfigurationContributor`, uses bytebuddy, the classloader needs to know the path to the bytebuddy.

In this changeset, similar to other dependencies, e.g. javapoet, finds bytebuddy jar file and adds it to the classpath to call `BootStrapCodeGenerator`.


snicoll (approved):

snicoll:
Merged by 1187b44

PR #972, opened 2021-08-14
Add codeGenDebugPort property to debug Spring AOT plugins

ttddyy:
When this parameter is present, launch the code gen process with remote debug mode.

Sample command:
```
> mvn -Pnative spring-aot:generate@generate -Dspring.aot.codeGenDebugPort=9000
```

### DIFF
--- a/spring-aot-gradle-plugin/src/main/java/org/springframework/aot/gradle/SpringAotGradlePlugin.java
+++ b/spring-aot-gradle-plugin/src/main/java/org/springframework/aot/gradle/SpringAotGradlePlugin.java
@@ -65,6 +65,8 @@ public class SpringAotGradlePlugin implements Plugin<Project> {
 
 	public static final String GENERATE_TEST_TASK_NAME = "generateTestAot";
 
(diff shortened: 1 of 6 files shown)


sdeleuze (changes requested):
Thanks a lot for this very interesting PR, any chance you could fix the typo and do the same on Grade side + add related documentation?

wilkinsona:
On the Gradle side, given that `GenerateAotSources` is already a `JavaExec` subclass, I wonder if it would be better to lean on its existing [debug options](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/JavaExec.html#debugOptions-org.gradle.api.Action-) for this, to some extent at least.

sdeleuze:
@wilkinsona Thanks for your feedback, could you explore this proposal @ttddyy ?

sdeleuze:
Given the fact we deeply need AOT debugging capabilities to move forward I will merge this PR as it is, and will create a related issue to implement @wilkinsona proposal.

ttddyy:
Thanks @wilkinsona  I'm still learning the gradle side, so appreciated the guidance.
I will address it and documentation in the new issue(#978)

Issue #978, opened 2021-08-18
Leverage Gradle debug options for codeGenDebugPort


sdeleuze:
As a follow-up of https://github.com/spring-projects-experimental/spring-native/pull/972, as proposed by @wilkinsona:

> On the Gradle side, given that `GenerateAotSources` is already a `JavaExec` subclass, I wonder if it would be better to lean on its existing [debug options](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/JavaExec.html#debugOptions-org.gradle.api.Action-) for this, to some extent at least.

Could you take care of that @ttddyy?

sdeleuze:
Fixed by d269be85932b10927ee8e7f1a683df66e8ef0ac4.

PR #979, opened 2021-08-19
Simplify "codeGenDebugPort" implementation in gradle plugin

ttddyy:
Previously, the `codeGenDebugPort` implementation in gradle plugin was
not effectively using the `JavaExec` infrastructure.  This commit
updates to use the debug options from `JavaExec` to configure the
remote debugging.

For code simplification, remove the `springAot` DSL support and only
take `codeGenDebugPort` project property. This is because, in most
cases, remote debug is specified in the command line.

Also, adding documentation about `codeGenDebugPort` parameter.


ttddyy:
For #978

sdeleuze:
Thanks @ttddyy for this contribution.

Issue #950, opened 2021-08-04
`repeated interface: org.springframework.aop.SpringProxy` with spring-security annotations on `@RestController`


ch4mpy:
I get following error as soon as I use both `@RequiredArgsConstructor` and `@PreAuthorize` in the same  a `@RestController` :
```
[INFO]     [creator]     Error: Error parsing dynamic proxy configuration in file:/workspace/META-INF/native-image/org.springframework.aot/spring-aot/proxy-config.json:
[INFO]     [creator]     java.lang.IllegalArgumentException: repeated interface: org.springframework.aop.SpringProxy
[INFO]     [creator]     Verify that the configuration matches the schema described in the -H:PrintFlags=+ output for option DynamicProxyConfigurationResources.
[INFO]     [creator]     com.oracle.svm.core.util.UserError$UserException: Error parsing dynamic proxy configuration in file:/workspace/META-INF/native-image/org.springframework.aot/spring-aot/proxy-config.json:
[INFO]     [creator]     java.lang.IllegalArgumentException: repeated interface: org.springframework.aop.SpringProxy
[INFO]     [creator]     Verify that the configuration matches the schema described in the -H:PrintFlags=+ output for option DynamicProxyConfigurationResources.
[INFO]     [creator]            at com.oracle.svm.core.util.UserError.abort(UserError.java:68)
[INFO]     [creator]            at com.oracle.svm.hosted.config.ConfigurationParserUtils.doParseAndRegister(ConfigurationParserUtils.java:139)
[INFO]     [creator]            at com.oracle.svm.hosted.config.ConfigurationParserUtils.lambda$parseAndRegisterConfigurations$3(ConfigurationParserUtils.java:120)
[INFO]     [creator]            at java.base/java.util.stream.ReferencePipeline$4$1.accept(ReferencePipeline.java:212)
[INFO]     [creator]            at com.oracle.svm.hosted.config.ConfigurationParserUtils$1.tryAdvance(ConfigurationParserUtils.java:109)
[INFO]     [creator]            at java.base/java.util.Spliterator.forEachRemaining(Spliterator.java:326)
[INFO]     [creator]            at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:658)
[INFO]     [creator]            at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:274)
[...]

PR #1002, opened 2021-08-26
gh-950 use Set instead of List in proxy descriptors

ch4mpy:
On Windows, it happens that `proxy-config.json` is generated with "repeated interface" when building native docker images (`spring-boot:build-image` with `<BP_NATIVE_IMAGE>true</BP_NATIVE_IMAGE>`)

Has spotted by @arielcarrera, this happens when a class is annotated with two annotations that require a proxy like @Validated and @ConfigurationProperties (in my case, it was a `@Controller` annotated with `@RestController`, `@RequiredArgsConstructor` and `@PreAuthorize`).

As I haven't unerstood the value of ordered collections inside proxy descriptors, I propose here to replace `List` by `Set` for private members and `Collection` in methods signatures.

This should fix https://github.com/spring-projects-experimental/spring-native/issues/950

### DIFF
--- a/spring-aot/src/main/java/org/springframework/nativex/domain/proxies/AotProxyDescriptor.java
+++ b/spring-aot/src/main/java/org/springframework/nativex/domain/proxies/AotProxyDescriptor.java
@@ -16,8 +16,9 @@
 
 package org.springframework.nativex.domain.proxies;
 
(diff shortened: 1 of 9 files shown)


ch4mpy:
Sorted collection must be used

PR #1011, opened 2021-09-01
Update AOT remote debug option in maven plugin

ttddyy:
Prior to this commit, the remote debugging for maven plugin used
`-Dspring.aot.codeGenDebugPort=<port>`.  This is convenient to specify
port but lacking the ability to specify different options.

This commit changes the parameter to `-Dspring.aot.debug` for maven
plugin.  This aligns with the widely used debug option style, for
example, surefire plugin.  When `-Dspring.aot.debug=true` is specified,
it uses port 5005 for remote debugging.  When othter than `true` is
specified, the value is passed to the AOT code gen process. So, users
can specify more advanced options.

cc @bclozel


bclozel:
Closed with 467b2ee8fac856e6

Thanks @ttddyy !

PR #1025, opened 2021-09-12
Add MariaDB hints

cschaible:
Hi, 
I've tested the hints in a two simple JPA applications against a local and a managed MariaDB in azure (by adding the hints directly to the main class) and it worked without problems.
I've tried to add a MariaDB configuration to the petstore-jpa sample in this repo but the build failed with connectivity issues to the docker database(java.sql.SQLException: null,  message from server: "Host '172.20.0.1' is not allowed to connect to this MySQL server"). 
Therefore I've dropped the changes in the sample application and just committed the hints.

### DIFF
--- a/spring-native-configuration/src/main/java/org/mariadb/MariaDbHints.java
+++ b/spring-native-configuration/src/main/java/org/mariadb/MariaDbHints.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2019-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


cschaible:
I've created a repo with a simple sample app where the hints are used (by adding the hints directly to the main class). The code can be found [here](https://github.com/cschaible/spring-native-mariadb-hints-sample/tree/master).

sdeleuze:
Thanks, merged via 5df5f1d6cc35c66f1aca97054d7762ad20555ebf, I modified your commit to reference the hint from `META-INF/services/org.springframework.nativex.type.NativeConfiguration`.

PR #1027, opened 2021-09-13
Start building against Spring boot 2.6.0 snapshots

snicoll:
This upgrades the build to Spring Boot 2.6.0 and Spring Cloud 2021.0.0 snapshots.

Some samples are broken following this upgrade:

* JPA and petclinic-jpa samples (and others Spring Data related) due to a reflection configuration missing for `PropertiesBasedNamedQueries` (that used to be package private). It's unclear why the constructor has to have a reflection entry considering we're not using reflection
* grpc, unrelated to the upgrade. It looks like the version of grpc we use is incompatible with the new generation

The events sample failed locally but I don't really understand why as running it again seems to work.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,8 +13,8 @@
 		<revision>0.11.0-SNAPSHOT</revision>
 		<graalvm.version>21.2.0</graalvm.version>
 		<docs.resources.version>0.2.1.RELEASE</docs.resources.version>
-		<spring.boot.version>2.5.5-SNAPSHOT</spring.boot.version>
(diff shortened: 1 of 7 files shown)


snicoll:
I confirm the JPA samples are also broken without this commit and are related to a recent change on `0.11.x` rather.

sdeleuze:
`grpc` works fine on my side, with or without `run-dev-container.sh`. Maybe a missing local `grpccurl` on your side.

sdeleuze:
Merged an updated version via 82e6c51050738c1f492b63d8c6ec19dce760f8aa.

PR #1057, opened 2021-09-21
Fix cloud-stream-kafka sample for 0.11.x

sobychacko:
With the changes in this PR, the sample in `cloud-stream-kafka` is working on `0.11.x` branch for the command: `build.sh --aot-only`. 

However, it fails for the non-aot mode build with `build.sh`.  Do you know why there is a difference between the two modes of build execution?


snicoll:
`build.sh` builds with AOT, create a native image and then check that the output is what's expected using `verifh.sh`.

`build.sh --aot-only` builds with AOT, start the app on the JVM and then check that the output is what's expected using `verifh.sh`.

So if it's failing with the former, some hints for GraalVM are missing? To be clear both executions uses the AOT engine (on native or on the JVM/hotspot respectively).

spring-native-configuration/src/main/java/org/springframework/cloud/stream/kafka/KafkaBinderHints.java
```diff
@@ -33,7 +34,7 @@
 						KafkaBinderHealthIndicatorConfiguration.class,
 						KafkaExtendedBindingProperties.class,
 						KafkaBindingProperties.class
-				})
+				}, access = AccessBits.FULL_REFLECTION)
```
sdeleuze, line 37: Could you please elaborate on why this change is required?

sdeleuze:
Closing this PR as cloud-stream-kafka sample is using dynamic application context and can't be supported on 0.11.

Issue #1048, opened 2021-09-21
Handle destroy callback


snicoll:
`CommonAnnotationBeanPostProcessor` is not invoked as we disable it when running in AOT mode. We should scan for `PostConstruct` and `PreDestroy` and update the bean definition metadata accordingly.

We're currently not writing that information so we should do that as well so that the container can pick it up. Since it's called via reflection, we'll have to update the native configuration for that as well.

snicoll:
Repurposing this issue to handle destroy callbacks as init callbacks are taken care of as part of #1059

snicoll:
Fixed by 1f96ae90

PR #1059, opened 2021-09-22
Handle init callbacks

christophstrobl:
This PR introduces support for detecting and calling bean initialization (post construct) methods.
Public, protected and package private init methods will be called directly, however private ones require addition reflection configuration.


spring-aot/src/main/java/org/springframework/context/bootstrap/generator/bean/descriptor/BeanInstanceDescriptor.java
```diff
@@ -47,12 +48,15 @@
 
 	private final List<PropertyDescriptor> properties;
 
+	private final List<InitializationCallback> initializationCallbacks;
```
snicoll, line 51: Having looked at the end result, I think this should rather be a `MemberDescriptor<Method>`. This way we can move the code generation to the writer and the reflection configuration can be provided based on the descriptor.

spring-aot/src/main/java/org/springframework/context/bootstrap/generator/bean/descriptor/BeanInstanceDescriptor.java
```diff
@@ -166,6 +178,37 @@ public CodeBlock write(String beanVariable) {
 		}
 	}
 
+	/**
+	 * Wraps the code that's necessary to invoce initialization (post construct) methods if any.
+	 */
+	public static class InitializationCallback {
```
snicoll, line 184: As a result this could go away.

spring-aot/src/main/java/org/springframework/context/bootstrap/generator/bean/descriptor/InitializationCallbacksSupplier.java
```diff
@@ -0,0 +1,106 @@
...
+
+	static {
+
+		EXTERNALLY_MANAGED_INIT_METHODS = ReflectionUtils.findField(RootBeanDefinition.class, "externallyManagedInitMethods");
+		ReflectionUtils.makeAccessible(EXTERNALLY_MANAGED_INIT_METHODS);
+	}
+
+	List<InitializationCallback> detectInstanceCallbacks(BeanDefinition beanDefinition) {
+
+		if(!(beanDefinition instanceof RootBeanDefinition)) {
```
snicoll, line 51: How about `getInitMethodName`? The idea is to honor all init method we know about. It should be this, and then any externally managed, if any.

spring-aot/src/main/java/org/springframework/context/bootstrap/generator/bean/descriptor/InitializationCallbacksSupplier.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Provide the {@link InitializationCallback initialization (post construct) callbacks} of a given class.
+ *
+ * @author Christoph Strobl
+ */
+class InitializationCallbacksSupplier {
+
+	private static final Field EXTERNALLY_MANAGED_INIT_METHODS;
+
+	static {
```
snicoll, line 43: I am not sure the optimization is worth it. We could have a method that would give us back the list based on a bean definition and hide all that reflection hackery there.

spring-aot/src/main/java/org/springframework/context/bootstrap/generator/bean/descriptor/InitializationCallbacksSupplier.java
```diff
@@ -0,0 +1,106 @@
...
+		RootBeanDefinition rootBeanDefinition = (RootBeanDefinition) beanDefinition;
+		Object field = ReflectionUtils.getField(EXTERNALLY_MANAGED_INIT_METHODS, rootBeanDefinition);
+
+		if (ObjectUtils.isEmpty(field)) {
+			return Collections.emptyList();
+		}
+
+		List<InitializationCallback> callbacks = new ArrayList<>();
+		Class<?> targetType = beanDefinition.getResolvableType().toClass();
+		for (Object initMethodName : (Iterable<? extends Object>) field) {
```
snicoll, line 64: We should rather put everything in a `LinkedHashSet` and iterate over it, to avoid cases where something is both detected by an initMethodName, and also annotated.

spring-aot/src/main/java/org/springframework/context/bootstrap/generator/bean/descriptor/InitializationCallbacksSupplier.java
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	List<InitializationCallback> detectInstanceCallbacks(Class<?> type) {
+
+		List<InitializationCallback> callbacks = new ArrayList<>();
+		ReflectionUtils.doWithMethods(type, method -> callbacks.add(getInitMethodCallback(method)), method -> !ObjectUtils.isEmpty(method.getAnnotationsByType(PostConstruct.class)));
+		return callbacks;
+	}
+
+	private InitializationCallback getInitMethodCallback(Method initMethod) {
```
snicoll, line 77: This code can move to `DefaultBeanInstanceSupplierWriter`.

spring-aot/src/main/java/org/springframework/context/bootstrap/generator/nativex/HintsBeanNativeConfigurationProcessor.java
```diff
@@ -44,16 +50,12 @@
 import org.springframework.nativex.type.TypeSystem;
 import org.springframework.util.ClassUtils;
 
-import static org.springframework.context.bootstrap.generator.infrastructure.nativex.NativeConfigurationRegistry.InitializationConfiguration;
```
snicoll, line 47: I know we don't have a consistent code style at the moment but can you please re-read the commit to avoid such unrelated changes?

spring-aot/src/main/java/org/springframework/context/bootstrap/generator/nativex/HintsBeanNativeConfigurationProcessor.java
```diff
@@ -62,6 +64,7 @@
 	@Override
 	public void process(BeanInstanceDescriptor descriptor, NativeConfigurationRegistry registry) {
 		findAndRegisterRelevantNativeHints(descriptor.getUserBeanClass(), registry);
+		registerReflectionForInitializtionMethodsIfNecessary(descriptor, registry);
```
snicoll, line 67: This is not the right place for this. This should go in `DefaultBeanNativeConfigurationProcessor`.

spring-aot/src/test/java/org/springframework/context/bootstrap/generator/bean/descriptor/InitializationCallbacksSupplierTests.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.context.bootstrap.generator.sample.callback.ConfigHavingBeansWithInitMethod.PrivateExternallyManagedInitMethod;
+import org.springframework.context.bootstrap.generator.sample.callback.ConfigHavingBeansWithInitMethod.PublicExternallyManagedInitMethod;
+import org.springframework.context.bootstrap.generator.test.CodeSnippet;
+
+/**
+ * Tests for {@link InitializationCallbacksSupplier}.
+ *
+ * @author Christoph Strobl
+ */
+class InitializationCallbacksSupplierTests {
```
snicoll, line 37: We need more tests here, in particular with `initMethodName`.

snicoll (commented):
Thanks for the PR. I've made a few comments. Looking at it a bit closed, I've realized what I've postponed should really happen now (injection of properties).

If we merge this, we'd call post construct before the properties are actually injected :(

snicoll:
With the help of @jhoeller, we'll be able to reduce reflection access once https://github.com/spring-projects/spring-framework/issues/27449 is done.

sdeleuze:
I removed the version assigned since #1048 also covers it.

snicoll:
I am going to review and merge this hopefully by M1. The other one is (partially) superseded by this one anyway.

snicoll:
Merged by 1f96ae90

Issue #739, opened 2021-04-15
Allow Netty reflection to fix RSS memory usage with Java 11


sdeleuze:
With Java 8 : 71M of RSS
With java 11 : 123M of RSS
Same figures with 21.0 or 21.1-dev.

cc @aclement @martinlippert

sdeleuze:
A part is not specific to Netty, see those figures for `commandlinerunner` with GraalVM 21.1.0.

**Java 8**
- Image build time: 22.5s
- RSS memory: 32.1M
- Image size: 24.1M

**Java 11**
- Image build time: 36.6s
- RSS memory: 37.0M
- Image size: 36.5M

**Java 16**
- Image build time: 40.8s
- RSS memory: 40.5M
- Image size: 39.7M

sdeleuze:
Figure after oracle/graal#3365 being fixed:

**GraalVM 21.1.0**
Build memory: 8.18GB
Image build time: 72.9s
RSS memory: 127.0M
Image size: 63.8M

**GraalVM 21.2.0-dev**
Build memory: 6.75GB
Image build time: 68.8s
RSS memory: 129.8M
Image size: 73.2M

So this seems a confirmation there is a Netty specific footprint issue on Java 11+.

sdeleuze:
Same results with GraalVM 21.2.0.

phxql:
I tracked this down to a behaviour in Netty. Netty on JDK <= 9 disables reflective access to a DirectByteBuffer constructor. Setting `-Dio.netty.tryReflectionSetAccessible=true` allows this again and fixes this issue.

See PR https://github.com/spring-projects-experimental/spring-native/pull/1073 for a fix.

martinlippert:
Thanks again for this fix, the results are quite impressive for our use case. The memory usage went down dramatically, absolutely amazing... 🥳

PR #1073, opened 2021-09-24
Allow netty reflection to fix RSS memory usage

phxql:
See #739

Setting -Dio.netty.tryReflectionSetAccessible=true configures Netty to use the DirectByteBuffer, which solves a bug where the Webflux Netty application used a lot more memory when built with GraalVM 11 vs GraalVM 8.


sdeleuze:
Merged via 289af3f43e09a9510a60451f6239b8fe16ab0667 thanks a lot @phxql!

Issue #1068, opened 2021-09-24
Fix Kafka samples


sdeleuze:
`kafka`, `kafka-avro` and `kafka-streams` samples currently broken at AOT level, cc @garyrussell.

PR #1078, opened 2021-09-24
GH-1068: Fix Kafka Hints and Samples

garyrussell:
Resolves https://github.com/spring-projects-experimental/spring-native/issues/1068

- change `@NativeHint` trigger
- narrowing generic type parameters from `Object` does not currently work with AOT

### DIFF
--- a/samples/kafka-avro/src/main/java/com/example/kafka/KafkaAvroApplication.java
+++ b/samples/kafka-avro/src/main/java/com/example/kafka/KafkaAvroApplication.java
@@ -35,17 +35,17 @@ public class KafkaAvroApplication {
 	}
 
 	@Bean
-	public ConcurrentMessageListenerContainer<String, Thing3> manualListenerContainer(MyMessageListener listener,
-			ConcurrentKafkaListenerContainerFactory<String, Thing3> factory) {
+	public ConcurrentMessageListenerContainer<Object, Object> manualListenerContainer(MyMessageListener listener,
+			ConcurrentKafkaListenerContainerFactory<Object, Object> factory) {
 
-		ConcurrentMessageListenerContainer<String, Thing3> container = factory.createContainer("graal");
+		ConcurrentMessageListenerContainer<Object, Object> container = factory.createContainer("graal");
 		container.getContainerProperties().setGroupId("graal3");
 		container.getContainerProperties().setMessageListener(listener);
 		return container;
 	}
 
 	@Bean
-	public ApplicationRunner runner(KafkaTemplate<String, Thing> template) {
+	public ApplicationRunner runner(KafkaTemplate<Object, Object> template) {
 		return args -> {
 			Thing thing = Thing.newBuilder().setStringField("someValue").setIntField(42).build();
 			template.send("graal", thing);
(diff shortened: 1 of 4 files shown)


spring-native-configuration/src/main/java/org/springframework/kafka/annotation/KafkaHints.java
```diff
@@ -69,7 +69,7 @@
 import io.confluent.kafka.serializers.KafkaAvroSerializer;
 import io.confluent.kafka.serializers.subject.TopicNameStrategy;
 
-@NativeHint(trigger=KafkaListenerConfigurationSelector.class,
+@NativeHint(trigger=KafkaListener.class,
```
sdeleuze, line 72: I think `KafkaListener` will be evaluated via a classpath check, could we use for example`KafkaListenerAnnotationBeanPostProcessor` instead which will be enabled only if such bean is active?

samples/kafka-avro/src/main/java/com/example/kafka/KafkaAvroApplication.java
```diff
@@ -35,17 +35,17 @@ public NewTopic topic() {
 	}
 
 	@Bean
-	public ConcurrentMessageListenerContainer<String, Thing3> manualListenerContainer(MyMessageListener listener,
-			ConcurrentKafkaListenerContainerFactory<String, Thing3> factory) {
+	public ConcurrentMessageListenerContainer<Object, Object> manualListenerContainer(MyMessageListener listener,
```
sdeleuze, line 38: I think we should not change the sample like that but instead fix the AOT processing to handle this use case cc @snicoll.

sdeleuze (changes requested):

samples/kafka-avro/src/main/java/com/example/kafka/KafkaAvroApplication.java
snicoll: Unfortunately, this isn't that easy. There's no `ConcurrentMessageListenerContainer<String, Thing3>` bean in the context so, technically, the current code is wrong. It is only wrong because Framework has several layers of fallback before throwing an exception that we don't have. 

I discussed this with @jhoeller and he was more keen to fix this rather than relying on the fallback. What bothers me is the fact we can't enforce that with the regular JVM/runtime so it always feel like a limitation of the AOT engine....

sdeleuze:
Merged via 453e1cbe8a0f2a050a59a9eb3c11600fcd031d70 with a `KafkaListenerAnnotationBeanPostProcessor` trigger, thanks!

samples/kafka-avro/src/main/java/com/example/kafka/KafkaAvroApplication.java
garyrussell: This is rather unfortunate; users have relied on this capability for years.

snicoll: I agree. Feel free to create a separate issue and we'll see what we can do.

PR #1096, opened 2021-09-28
Make configserver sample work with Spring AOT.

OlgaMaciaszek:


sdeleuze:
Merged without disabling the compatibility verification

PR #1099, opened 2021-09-29
First pass at modifying SpringApplication on the fly

aclement:
This removes `SpringApplication.java` from the spring-native project. Instead it includes a contributor that rewrites the bytecode of the SpringApplication visible on the class path and puts it into the output folder of the project. If patches cannot be made that are expected to be made, an exception will be thrown.


sdeleuze:
Awesome, merged with 2 refinements:
 - Fix for the GRPC sample 3156c4691b442dc4b03812b007079fe1c3183569
 - Takes in account modifications from 538faacdfc56a67637a509440f128ea37ff17937

I have created the follow-up #1102 issue.

PR #1106, opened 2021-09-29
Reduce eureka client hints

OlgaMaciaszek:


spring-native-configuration/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientHints.java
```diff
@@ -30,6 +30,7 @@
 import org.springframework.nativex.hint.TypeHint;
 import org.springframework.nativex.type.NativeConfiguration;
 
+// TODO Not sure AccessBits.ALL is needed on these, to be checked
 @NativeHint(trigger = LoadBalancerEurekaAutoConfiguration.class, types = {
```
sdeleuze: Did you try without those hints on `EurekaLoadBalancerClientConfiguration` and `EurekaReactiveDiscoveryClientConfiguration`? In theory that should not be needed.

spring-native-configuration/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientHints.java
```diff
@@ -30,6 +30,7 @@
 import org.springframework.nativex.hint.TypeHint;
 import org.springframework.nativex.type.NativeConfiguration;
 
+// TODO Not sure AccessBits.ALL is needed on these, to be checked
```
sdeleuze: Could we use this opportunity to check if `AccessBits.ALL` is really needed or not? Something you can probably test is to remove `, access = AccessBits.ALL` on the hints with `EurekaClientAutoConfiguration` to just get the default reflection at class and public constructor level. If more is needed, could be great to add in a more fine grained fashion.

sdeleuze (changes requested):

PR #1107, opened 2021-09-29
Add note on Config Client support.

OlgaMaciaszek:


spring-native-docs/src/main/asciidoc/support.adoc
```diff
@@ -102,6 +102,8 @@ NOTE: When using Spring Native, `spring.cloud.refresh.enabled` is set to `false`
 
 NOTE: Spring Cloud Bootstrap is no longer supported.
 
+NOTE: While building a project that contains Spring Cloud Config Client, it is necessary to make sure that the configuration data source that it connects to (such as, Spring Cloud Config Server, Consul, Zookeeper, Vault, etc.) is available. For example, if you retrieve configuration data from Spring Cloud Config Server, make sure you have its instance running and available at the port indicated in the Config Client setup.
```
snicoll, line 105: Can we explain why that's necessary, something like "The application context is optimized at build time and requires the target environment to be resolved" or something like that.

sdeleuze (changes requested):
Could you please provide more context as suggested by Stéphane?

spring-native-docs/src/main/asciidoc/support.adoc
OlgaMaciaszek: Sure. will add.

PR #1131, opened 2021-10-05
Add lazy bean support

ttddyy:
Support `@Lazy` beans.

Signed-off-by: Tadaya Tsuyukubo <tadaya@ttddyy.net>


snicoll (approved):

snicoll:
If we merge this, a proxy is going to be created and I am not sure it would be in native so we probably need a sample that uses this feature to provide it works in both worlds.

@ttddyy would you like to give this a try? You could add that to `boot-features` or you could create a `framework-features` sample rather.

ttddyy:
Yes, having a sample project sounds better. I'll try to create one and probably mix with `ObjectProvider` usage as well.

> If we merge this, a proxy is going to be created and I am not sure it would be in native so we probably need a sample that uses this feature to prove it works in both worlds.

My understanding of lazy beans is that they skip the eager initialize at `appContext#refresh` [here](https://github.com/spring-projects/spring-framework/blob/main/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java#L913-L928), and get initialized at the first access. So, not sure, it involves any proxies.

ttddyy:
@snicoll I have added lazy bean usage to the existing `features` sample.

Issue #1155, opened 2021-10-18
Add support for FactoryBean constructor


artembilan:
There are some use-cases when `FactoryBean` target type cannot be determined from generics or some other info during `beanFactory.resolveDependency()`.
For this purpose we rely on the `FactoryBean.OBJECT_TYPE_ATTRIBUTE` attribute setting on `BeanDefinition`.
Such an attribute should be copied to the target generated code.
This way we won't need to develop our own simple `BeanRegistrationWriter` just for a sake of this attribute carrying on.

snicoll:
I changed my mind and I added a dedicated writer for the case of a `FactoryBean` whose target type is not defined upfront but via the attribute. It does some additional checks to verify that the attribute is compatible with the bounds of the factory bean.

Issue #1134, opened 2021-10-05
Restore Spring Integration support


sdeleuze:
To be discussed with @artembilan and @snicoll.

sdeleuze:
Awesome, thanks a lot @snicoll.

sdeleuze:
I am moving this one to 0.11.0 in order to allow us to focus on testing which is the highest priority.

snicoll:
Spring Integration used to work on AOT and now doesn't anymore so I think we need to make sure we fix that first by not changing things in Spring Integration (or fixing them if they have to be).

The sample fails currently with:

```
2021-11-08 09:54:19.410  INFO 13418 --- [           main] trationDelegate$BeanPostProcessorChecker : Bean 'org.springframework.integration.config.IntegrationManagementConfiguration' of type [org.springframework.integration.config.IntegrationManagementConfiguration] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
2021-11-08 09:54:19.430  INFO 13418 --- [           main] trationDelegate$BeanPostProcessorChecker : Bean 'integrationChannelResolver' of type [org.springframework.integration.support.channel.BeanFactoryChannelResolver] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
2021-11-08 09:54:19.432  INFO 13418 --- [           main] trationDelegate$BeanPostProcessorChecker : Bean 'integrationDisposableAutoCreatedBeans' of type [org.springframework.integration.config.annotation.Disposables] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
2021-11-08 09:54:19.671  INFO 13418 --- [           main] o.s.j.d.e.EmbeddedDatabaseFactory        : Starting embedded database: url='jdbc:h2:mem:d748fea7-eb9f-4316-817f-59fd975a0704;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=false', username='sa'
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.springframework.util.ReflectionUtils (jar:file:/Users/snicoll/workspace/work/experimental/spring-native/samples/integration/target/integration-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-core-5.3.13-SNAPSHOT.jar!/) to constructor java.lang.invoke.MethodHandles$Lookup(java.lang.Class)
[...]

PR #1172, opened 2021-10-22
Switch sample to Spring Integration 5.5.6-SNAPSHOT

artembilan:
Fixes https://github.com/spring-projects-experimental/spring-native/issues/1155
Related to https://github.com/spring-projects-experimental/spring-native/issues/1134

The `FactoryBean.OBJECT_TYPE_ATTRIBUTE` is crucial for those `FactoryBean` impls
which does not expose the target type during compilation.

* Add `FactoryBean.OBJECT_TYPE_ATTRIBUTE` as a candidate for the
`DefaultBeanRegistrationWriter.getAttributeFilter()` since that one is `false` by default anyway
* Rework `IntegrationApplication` to reflect the current Spring Boot state
* Use `spring-integration-5.5.6-SNAPSHOT` for the latest bean definition changes over there

The `build.sh -a` now passes for the integration sample.
Cannot confirm in the native mode since GraalVM doesn't work well on Windows: too long command line.
Need to build classpath file (jar with respective META-INF) somehow...


snicoll:
Thanks Artem. With your help I was able to move further and inspect the problematic bean definition. I can see that the bean definition for the `integrationApplication$ControlBusGateway` has a `FactoryBean.OBJECT_TYPE_ATTRIBUTE` but it doesn't seem to have the right type to me and I am wondering why writing it back works the way it does.

AFAIK this attribute should either be a `Class` or a `ResolvableType`. Any other type is ignored so I guess the current `org.springframework.beans.BeanMetadataAttribute` is converted somewhere to a class. This explains why me checking for the attribute didn't work as I didn't expect such a type.

snicoll:
> Cannot confirm in the native mode since GraalVM doesn't work well on Windows: too long command line.
Need to build classpath file (jar with respective META-INF) somehow...

It does not work but I am making progress:

```Error creating bean with name 'webFluxIntegrationRequestMappingHandlerMapping': Error setting property values; nested exception is org.springframework.beans.NotWritablePropertyException: Invalid property 'order' of bean class [org.springframework.integration.webflux.inbound.WebFluxIntegrationRequestMappingHandlerMapping]: Bean property 'order' is not writable or has an invalid setter method. Does the parameter type of the setter match the return type of the getter?```

**Comment by snicoll** [comment]
So the problem here is that we don't call `BeanDefinitionRegistryPostProcessor` that are defined as bean definitions. It's a tricky problem because if we did, we should remove the bean definition for it without any guarantee whatsoever that the bean is not used for other purposes. 

What's happening now is that a BDRPP defined in Spring Integration is invoked at runtime, and create the `webFluxIntegrationRequestMappingHandlerMapping` bean definition only at runtime which has not been procssed for native use.

**Comment by artembilan** [comment]
Thanks, Stephane, for feedback!
I wish I could run GraalVM build on my Windows to see these problems myself...

Nevertheless. Spring Integration has only this `BDRPP` - `IntegrationConfigurationBeanFactoryPostProcessor`, which is registered only once when an `@EnableIntegration` is processed.
That one does this logic:
```
List<IntegrationConfigurationInitializer> initializers =
				SpringFactoriesLoader.loadFactories(IntegrationConfigurationInitializer.class,
						beanFactory.getBeanClassLoader());

for (IntegrationConfigurationInitializer initializer : initializers) {
	initializer.initialize(beanFactory);
}
```

So, yes, it is indeed a programmatic bean definition registrations.

I guess, I'll revise my `IntegrationConfigurationInitializer` implementations in favor of just `instanceSupplier` (no `addPropertyValue()` or so), which is still supported in Native runtime I guess.

snicoll:
There's no need to do anything Artem. I am working on this already and have made progress, see https://github.com/spring-projects-experimental/spring-native/issues/1213

Issue #1202, opened 2021-11-01
Upgrade to native build tools 0.9.7.1


sdeleuze:
Spring Boot 2.6.0-RC1 is based on JUnit 5.8.1, so it should be fine.

I did a quick try, upgrading to 0.9.7.1, adding `<extensions>true</extensions>` in Maven `pom.xml` and removing `junit-platform-native` dependency, but I got the following failure message.

> Test configuration file wasn't found. Make sure that test execution wasn't skipped.

As pointed out in https://github.com/spring-projects-experimental/spring-native/issues/1192#issuecomment-956214014, `org.junit.platform.launcher.TestIdentifier` build-time initialization configuration should be removed from `SpringBootTestHints`.

Could you please take care of that @sbrannen?

sbrannen:
> Test configuration file wasn't found. Make sure that test execution wasn't skipped.

This happens because the version of Maven Surefire in use is using an older version of the `junit-platform-launcher` artifact that does not contain the `UniqueIdTrackingListener` that was introduced in JUnit 5.8 (JUnit Platform 1.8).

There are two ways to _fix_ this.

1. Declare an explicit test-scoped dependency on the `junit-platform-launcher` artifact (see below).
2. Upgrade to Maven Surefire 3.0.0-M5 as used in the sample applications in the Native Build Tools project.

sbrannen:
Note that the auto-alignment of JUnit Platform versions was added in Maven Surefire 3.0.0-M4 (https://issues.apache.org/jira/browse/SUREFIRE-1585).

sbrannen:
Superseded by #1206

PR #1206, opened 2021-11-02
Upgrade to Native Build Tools 0.9.7.1

sbrannen:
See #1202 for background information.

----

This PR upgrades the `spring-aot-gradle-plugin` and all Maven based sample apps to Native Build Tools (NBT) 0.9.7.1

NBT 0.9.6 requires JUnit Platform 1.8 (JUnit 5.8) due to its dependency on the `UniqueIdTrackingListener` in the `junit-platform-launcher` artifact. The Maven sample apps are currently based on Spring Boot 2.6 RC1 which uses JUnit 5.8.1 (JUnit Platform 1.8.1). Thus, there is no conflict between the JUnit version used in the sample apps and the version used by NBT.

However, there is an issue regarding Maven Surefire.

Maven Surefire 3.0.0-M4 introduced support for automatically aligning the JUnit Platform version used by Surefire with the version needed by the user's configured version of JUnit Jupiter. Since we are not using Surefire 3.0 milestones, the parent POM for the Maven sample apps has to explicitly declare a dependency on the correct version of the `junit-platform-launcher` in order to make the `UniqueIdTrackingListener` available for Native Build Tools 0.9.6 and higher. The actual version of the `junit-platform-launcher` is managed by the junit-bom that is imported by the `spring-boot-starter-parent`.

This PR also removes build-time initialization for `org.junit.platform.launcher.TestIdentifier`, since that is already handled by the `JUnitPlatformFeature` in the Native Build Tools (starting with version 0.9.6).

### DIFF
--- a/samples/maven-parent/pom.xml
+++ b/samples/maven-parent/pom.xml
@@ -66,6 +66,7 @@
 					<groupId>org.graalvm.buildtools</groupId>
 					<artifactId>native-maven-plugin</artifactId>
 					<version>${native.buildtools.version}</version>
+					<extensions>true</extensions>
 					<executions>
 						<execution>
 							<goals>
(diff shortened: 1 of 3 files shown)


sdeleuze:
@sbrannen Could you please also update the documentation accordingly (version, Maven configuration)?

sdeleuze (changes requested):
Could you please also update the documentation accordingly (version, Maven configuration)?

sdeleuze:
I will do the polish myself in order to be able to run a CI full build tonight.

sbrannen:
> I will do the polish myself in order to be able to run a CI full build tonight.

sbrannen:
> Could you please also update the documentation accordingly (version, Maven configuration)?

Regarding the Maven configuration, I think we should probably add a note to the user documentation similar to the following comment:

https://github.com/spring-projects-experimental/spring-native/blob/41591f840429267722fcc0f78bc6ad9648480b7a/samples/maven-parent/pom.xml#L183-L193

Otherwise, Spring Boot users will be confused when the Native Build Tools Maven plugin fails to find the "unique ID files" that should be generated by the `UniqueIdTrackingListener`.

@sdeleuze, what do you think?

sdeleuze:
@sbrannen Good point indeed, could you please update https://github.com/spring-projects-experimental/spring-native/blob/main/spring-native-docs/src/main/asciidoc/getting-started-native-build-tools.adoc accordingly?

PR #1214, opened 2021-11-04
Transition JpaComponentProcessor to BeanFactoryNativeConfigurationProcessor

christophstrobl:
Replace the `JpaComponentProcessor` with a `BeanFactoryNativeConfigurationProcessor` implementation.

Resolves: #1208


sdeleuze (changes requested):
I tested this PR locally, I see a lot of logs from `org.springframework.data.JpaConfigurationProcessor` and very slow test execution when running `./build.sh` at the root. Could you please double check if something need to be refined?

christophstrobl:
This might be caused by multiple scans of the path to looking for types annotated with `javax.persistence.Entity`.

sdeleuze:
Just when running tests it seemed to be significantly pretty slow, so I am wondering if an optimization could be done to avoid too much performance impact.

sdeleuze:
Merged thanks!

PR #1238, opened 2021-11-08
Move Kafka ComponentProcessor to AOT form

garyrussell:
Beans that are Kafka listeners can have parameters with Avro generated
types. Such types are annotated with `@AvroGenerated`.

These classes need to be added to the reflection configuration.

There are two generic types for the record key and value; each of which
could be an Avro type.

Given 2 Avro types `Thing1` and `Thing2`, there are several types of
listener that receive these types.

Implementation of `MessageListener` or `BatchMessageListener`.

```
class MyListener implements MessageListener<Thing1, Thing2> { ... }
```

In this case, we need to use `ResolvableType` to get the generic types.

```
@KafkaListener
void listen1(Thing1 thing) { ... }
```

```
@KafkaListener
void listen2(List<Thing1> things) { ... }
```

```
@KafkaListener
void listen(ConsumerRecord<Thing1, Thing2> record)) { ... }
```

```
@KafkaListener
void listen(ConsumerRecords<Thing1, Thing2> record)) { ... }
```

In each of these cases detect if the parameters are `AvroGenerated` and add reflection
hints.

Also improve sample to actually use all 3 avro types.


sdeleuze:
Merge via 95b450a284aa0cae97c413b7c48031617135a103, thanks.

PR #1259, opened 2021-11-14
avoid NPE and improve log message, default application class if missing

hdeadman:
I was specifying mainClass in the build (b/c my main class is in a dependency jar) but it was not getting used in the `generateAotTest` task of the build so this adds a null check. 

The private `findMethod` looking for destroy methods was confused by a FactoryBean that was explicitly setting a destroy method and then not finding it on the `FactoryBean` interface. In my case the `FactoryBean` returned a `DisposableBean` so it didn't need to set the destroy method, but this change adds the bean definition to the exception thrown so it is easier to identify where the problem is. 

This also defaults applicationClass to mainClass if applicationClass isn't discovered. Maybe applicationClass should be configurable like mainClass but even if it were, mainClass might be a reasonable default? If that isn't the case I can look at making applicationClass configurable via the plugins.

### DIFF
--- a/spring-aot/src/main/java/org/springframework/aot/build/ApplicationStructure.java
+++ b/spring-aot/src/main/java/org/springframework/aot/build/ApplicationStructure.java
@@ -64,7 +64,8 @@ public class ApplicationStructure {
 		this.classesPaths = classesPaths;
 		this.mainClass = mainClass != null ? mainClass : detectMainClass(classesPaths);
 		this.testClasses = testClasses;
-		this.applicationClass = detectApplicationClass(classesPaths);
+		String applicationClass = detectApplicationClass(classesPaths);
+		this.applicationClass = applicationClass == null ? mainClass : applicationClass;
 		this.classpath = classpath;
 		this.classLoader = classLoader;
 	}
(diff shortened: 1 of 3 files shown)


sdeleuze:
> This also defaults applicationClass to mainClass if applicationClass isn't discovered. Maybe applicationClass should be configurable like mainClass but even if it were, mainClass might be a reasonable default? If that isn't the case I can look at making applicationClass configurable via the plugins.

@snicoll Related to that, we currently only search application class on sources not JARs since I was wondering about potential side effects and performance impact. I am wondering if we should search in dependencies as well, maybe we should do what Spring Boot does for main class detection?

snicoll:
@hdeadman there seem to be two completely different changes in this PR and we'd rather have one change at a time. Please move the `findMethod` method to a separate PR and we'd need a test that showcases the factorybean problem you were describing.

sdeleuze:
I am working on a refined fix for the application class issue.

sdeleuze:
I fixed the application context class customization via 0673e09a9456612e67d216ed43a6270a2a076724.

PR #1260, opened 2021-11-14
Add flyway hints

cschaible:
This commit contains the adjustments that were made in micronaut and
quarkus to use flyway in native binaries.

The files were copied/took over from the micronaut repository which
copied/took over the files from quarkus.

A minor adjustment was made to use the flyways logger factory instead of
slf4j.

The micronaut repository with the sources can be found here:
https://github.com/micronaut-projects/micronaut-flyway/

The quarkus repository with the sources can be found here:
https://github.com/quarkusio/quarkus/tree/main/extensions/flyway/

The location of flyway scripts can be customized by adding the
flyway.locations property to the buildArgs of the native build
maven/gradle task. It is recognized at build time:

```
nativeBuild {
  ...
  buildArgs.set(listOf(
    "-Dflyway.locations=classpath:db/migration/postgres"))
}
```

---
I am not sure if the location of the files is correct as it contains classes with
@Substitute/@Alias annotations. I haven't found them in the spring-native-configuration repository.
I haven't tested the files after copying them from my test project to this repo. Is it possible to use the build results of the PR build to test the change in a demo application? Or is there another way of testing it without configuring the maven-publish plugin in the spring-native-configuration repository?

---
I have copied the files from micronaut and did small adjustments as described above.
I've removed the `@author` and `@since` annotations from the javadoc as this relates probably mainly to the micronaut changes (at least partially they are identical to the ones in the quarkus repository). I'm not sure whether they have to be kept or whether it's sufficient to have the links to the files in the other repos in the javadoc.

Additionally I had the impression that the latest quarkus version is a little bit different. With the latest changes from the quarkus repo i couldn't start the application successfully (initially they probably had something more similar to the micronaut solution).
The changes that were taken over from micronaut are working successfully in my sample application.

The sample application can be found here:
https://github.com/cschaible/spring-native-intro-samples/tree/master/csc.native.jpa

The docker containers can be started by running the `docker/up.sh` script.
The application can be built by running `./gradlew clean nativeBuild`.
The binary can be started by running: `SPRING_PROFILES_ACTIVE=postgres ./csc.native.jpa`.

---
Please add comments if you have refactoring wishes or feel free to adjust the code yourself.

### DIFF
--- a/spring-native-configuration/pom.xml
+++ b/spring-native-configuration/pom.xml
@@ -158,6 +158,11 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.flywaydb</groupId>
+            <artifactId>flyway-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>org.springframework.experimental</groupId>
             <artifactId>spring-aot</artifactId>
         </dependency>
--- a/spring-native-configuration/src/main/java/org/flywaydb/FeatureDetectorSubstitutions.java
+++ b/spring-native-configuration/src/main/java/org/flywaydb/FeatureDetectorSubstitutions.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2019-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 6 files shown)


sdeleuze:
Thanks for this initial work, but for this kind of contribution with code coming from various frameworks, a better home will be the GraalVM native configuration repository + asking agreement of the original authors would be better IMO. We will share more about how to contribute to this repository shortly, stay tuned. Also be aware that tests will have to be provided.

PR #1263, opened 2021-11-16
Provide more context when InitDestroyMethodsDiscoverer fails to discover a defined method

hdeadman:
Here is an attempt to recreate the problem I ran into and added the extra message info in this [PR](https://github.com/spring-projects-experimental/spring-native/pull/1259)

These were the beans I ran into:
```
    @ConditionalOnMissingBean(name = "httpClient")
    @Bean(destroyMethod = "destroy")
    public FactoryBean<SimpleHttpClient> httpClient() throws Exception {
        return buildHttpClientFactoryBean();
    }

    @ConditionalOnMissingBean(name = "noRedirectHttpClient")
    @Bean(destroyMethod = "destroy")
    public HttpClient noRedirectHttpClient() throws Exception {
        return getHttpClient(false);
    }

    @ConditionalOnMissingBean(name = "supportsTrustStoreSslSocketFactoryHttpClient")
    @Bean(destroyMethod = "destroy")
    public HttpClient supportsTrustStoreSslSocketFactoryHttpClient() throws Exception {
        return getHttpClient(true);
    }
```
from [here](https://github.com/apereo/cas/blob/0b9113c562c5912eb1860075e8f594d251dd0d31/core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreHttpConfiguration.java#L79)

These may be an odd case b/c they explicitly are setting a destroy method even though SimpleHttpClient (impl of HttpClient) and the SimpleHttpClientFactoryBean returned from these bean methods implement DisposableBean interface so they don't need to set destroy method. That has since been fixed and spring-native isn't erroring out anymore but I am not sure it is finding the destroy methods for those beans.

When spring-native processed the first bean the error message was:
```
java.lang.IllegalStateException: Lifecycle method annotation 'destroy' not found on: interface org.springframework.beans.factory.FactoryBean
```
and something similar about not finding `destroy` method on the `HttpClient` interface for the other two beans. 

The error didn't really point me to the bean causing the problem so I modified the exception to include the bean definition information in the exception message. 

You can see that message generated by the `processDisposableFactoryBean` test method in this PR. Both tests pass if the commented out `beanDefinition.setTargetType()` method is called.  My examples used FactoryBean but maybe this issue is just any Bean with an interface that doesn't contain the destroy method in it (b/c other interfaces are also implemented that do have that method) and since the bean explicitly defines a destroy method in the bean definition, spring-native looks for method on the interface defined as the bean's type. 

I don't know if this needs a fix or not, and feel free to delete this PR, I just thought the extra info in the exception message was useful for tracking down the bean causing the problem. I can remove the extra message from my other PR or move it to its own PR.

### DIFF
--- a/spring-aot/src/test/java/org/springframework/aot/context/bootstrap/generator/infrastructure/InitDestroyMethodsDiscovererTests.java
+++ b/spring-aot/src/test/java/org/springframework/aot/context/bootstrap/generator/infrastructure/InitDestroyMethodsDiscovererTests.java
@@ -26,6 +26,7 @@ import org.junit.jupiter.api.Test;
 import org.springframework.aot.context.bootstrap.generator.infrastructure.nativex.DefaultNativeReflectionEntry;
 import org.springframework.aot.context.bootstrap.generator.infrastructure.nativex.NativeConfigurationRegistry;
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.FactoryBean;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.support.DefaultListableBeanFactory;
 import org.springframework.beans.factory.support.RootBeanDefinition;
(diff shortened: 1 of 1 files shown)


snicoll:
@hdeadman thank you very much for doing this!

Looking at `processDisposableFactoryBean`, the current outcome looks fine to me. You are creating a bean whose type is exposed as `FactoryBean`. We don't create instance so we can't discover what the actual instance type is going to be. The exception message should certainly contain the bean name in this case, I think.

The second example is a bug. It should check for the destroy method on the target bean, not the factory bean.

I'll make use of your contribution to improve the error report. Thanks very much!

snicoll:
I've improved the error message based on your feedback. See #1280 for the second scenario exposed in this PR. Thanks again.

Issue #1264, opened 2021-11-16
Dynamically replace DITEL with AotDependencyInjectionTestExecutionListener


sbrannen:
## Overview

This is a follow up to #1258.

## Deliverables

- [x] Dynamically replace the standard `DependencyInjectionTestExecutionListener` with the `AotDependencyInjectionTestExecutionListener`

sbrannen:
Current work on this issue can be viewed in the following feature branch.

https://github.com/spring-projects-experimental/spring-native/compare/main...sbrannen:issues/gh-1264-revise-AOT-DITEL

PR #1265, opened 2021-11-16
Dynamically replace DITEL with AotDependencyInjectionTestExecutionListener

sbrannen:
Prior to this commit, the AotDependencyInjectionTestExecutionListener
(AOT_DITEL) was registered via spring.factories. In addition, the
AOT_DITEL was registered immediately before the standard
DependencyInjectionTestExecutionListener (DITEL), based on their
respective orders 1999 and 2000.

With this commit, the AOT_DITEL now extends DITEL and is registered
programmatically via the TestExecutionListenerFactoriesCodeContributor
to replace the DITEL registered via spring.factories in spring-test. In
addition, AOT_DITEL delegates to the standard behavior in DITEL if the
current test class is not supported via an AOT generated application
context, thereby making the AOT_DITEL a drop-in replacement for the
DITEL.

This has the following consequences.

- The AOT_DITEL is now registered instead of the DITEL, which means
  that there is only one "dependency injection" TestExecutionListener
  registered at any given time.
- The SpringBootDependencyInjectionTestExecutionListener is no longer
  registered by its DefaultTestExecutionListenersPostProcessor, since
  there is no longer a DITEL that it can replace.

See gh-1264

----

```
GraalVM: GraalVM 21.3.0 Java 11 CE (Java Version 11.0.13+7-jvmci-21.3-b05)
Date           Sample             Build Time (s)  Build Mem (GB)  RSS Mem (M)  Image Size (M)  Startup Time (s)  JVM Uptime (s)   ReflectConfig (lines)
20211116-1942  commandlinerunner  41.8            4.96            21.4         23.4            0.016             0.021           640
20211116-1944  webflux-netty      83.3            6.64            42.8         56.9            0.047             0.049           2346
20211116-1945  webmvc-tomcat      80.4            6.44            47.4         53.6            0.068             0.07            2410
20211116-1947  webflux-thymeleaf  93.4            7.16            51.1         60.9            0.044             0.045           2655
20211116-1952  grpc               48.6            5.26            26.5         26.0                                              360
20211116-1954  jdbc-tx            99.5            5.84            61.4         64.5            0.066             0.069           2964
20211116-1956  class-proxies-aop  90.7            6.20            59.0         62.3            0.074             0.077           2535
20211116-1958  batch              85.4            5.69            62.2         52.0            0.055             0.057           1693
```

### DIFF
--- a/spring-aot/src/main/java/org/springframework/aot/factories/TestExecutionListenerFactoriesCodeContributor.java
+++ b/spring-aot/src/main/java/org/springframework/aot/factories/TestExecutionListenerFactoriesCodeContributor.java
@@ -25,9 +25,14 @@ import org.springframework.nativex.domain.reflect.MethodDescriptor;
 /**
  * {@link FactoriesCodeContributor} that contributes source code for
  * {@link org.springframework.test.context.TestExecutionListener} implementations.
+ *
  * <p>Instead of instantiating them statically, we make sure that
  * {@link org.springframework.core.io.support.SpringFactoriesLoader#loadFactoryNames(Class, ClassLoader)}
  * will return their names and that reflection metadata is registered for native images.
+ *
+ * <p>In addition, this {@code FactoriesCodeContributor} transparently replaces the
+ * {@code DependencyInjectionTestExecutionListener} from {@code spring-test} with
+ * the {@code AotDependencyInjectionTestExecutionListener} from {@code spring-native}.
  *
  * @author Sam Brannen
  * @author Brian Clozel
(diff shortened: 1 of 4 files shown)


sdeleuze (changes requested):
This PR seems to break `commandlinerunner-gradle` tests:

```
> Task :test FAILED

CommandlinerunnerApplicationTests > autowiringWorks() FAILED
    java.lang.AssertionError at CommandlinerunnerApplicationTests.java:32
```
Could you please check and fix if confirmed?

sbrannen:
> This PR seems to break `commandlinerunner-gradle` tests:

The PR reveals an existing bug in the Gradle plugin.

Specifically, the generated `*-aot-test.jar` is appended to the classpath instead of prepended to it.

The result is that the generated `SpringFactoriesLoader` appears on the classpath after the standard `SpringFactoriesLoader` present in the `spring-core` JAR.

I have a fix in place locally.

sbrannen:
> Could you please check and fix if confirmed?

I fixed the bug in the Gradle plugin in eba6bc43b53ee3fb07f9378064f67af4641ea556, rebased on `main`, and force pushed the changes.

PR #1296, opened 2021-11-24
Data processors should inspect declared classes if present.

christophstrobl:
This PR makes sure declared classes are covered by the type inspection of data processors.


aclement (approved):
I deactivated the resources handler code and the data-jpa sample was failing (due to the SomeEnum inner enum inside the Foo class). I then applied your branch and it started working. So I think this is doing what we want. I haven't run it over all the samples but I imagine, since it is additive it won't break things.

PR #1311, opened 2021-11-29
GH-1134: Restore Spring Integration support

artembilan:
Fixes https://github.com/spring-projects-experimental/spring-native/issues/1134

* Upgrade sample to the latest Spring Integration `5.5.7-SNAPSHOT`
* Add `maven-shade-plugin` and `native-maven-plugin` under Windows profile
to make native image build working on Windows.
Se more info in Native Build Tools docs:
https://graalvm.github.io/native-build-tools/latest/maven-plugin.html#long_classpath_and_shading_support
This profile simply can go to the `../maven-parent/pom.xml` to make all the samples working on Windows.
Or you can remove this on merge since it might be just a noise for a broader community
* Add custom `ReactorClientHttpConnector` bean into the sample app, since by default a DNS resolver
fails with NPE on Windows in native image.
See more info in the: https://github.com/spring-projects/spring-framework/issues/27749
* Add more type nints into the `IntegrationHints`

### DIFF
--- a/samples/integration/pom.xml
+++ b/samples/integration/pom.xml
@@ -17,6 +17,7 @@
 
 	<properties>
 		<java.version>11</java.version>
+		<spring-integration.version>5.5.7-SNAPSHOT</spring-integration.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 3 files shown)


sdeleuze:
Since @artembilan, is Spring Integration `5.5.7-SNAPSHOT` required or are we fine with `5.5.6` shipped with Spring Boot 2.6.1 and used with Spring Native 0.11.0. If `5.5.7-SNAPSHOT` I think I will move integration support to Spring Native `0.11.1` to be released in a few weeks.

artembilan:
Thanks, @sdeleuze , for review!

Well, even if Spring Boot `2.6.1` was not planned ahead to let us slip into with our releases as well, and even if Spring Boot `2.6.2` is still not planned, that fully doesn't mean we cannot merge this:

1. The fix for `IntegrationHints` is version independent.
2. The fix for sample contains proper version.
3. When we merge this and give it end-users, what they only would need to do is to have the same `<spring-integration.version>5.5.7-SNAPSHOT</spring-integration.version>` in their projects for time being.

Not sure if you have anything else in project to say what is supported and what is not, but I don't see obstacles to have this fix in the upcoming `0.11.0` release.

sdeleuze:
Yeah I agree, I will merge a subset of this PR, thanks!

sdeleuze:
Merged via 13272e9c647db2b806e58136c567f5090c72d036, good progresses!

sdeleuze:
Feedback on the elements not integrated:
 - I did 't want to integrate the Netty update before having a feedback on https://github.com/spring-projects/spring-framework/issues/27749. Rossen has since provided an update to this point which will be fixed on Windows with Spring Boot 2.6.2 I guess, so Spring Native 0.11.1.
 - We should improve Windows support but via a distinct and dedicated issue for all samples with related doc, etc. See https://github.com/spring-projects-experimental/spring-native/issues/1319 I created for that purpose.

Issue #1262, opened 2021-11-15
Fix MergedContextConfiguration cache key for runtime contexts


snicoll:
While #1242 prevents certain `ContextCustomizerFactory` implementations from running again at runtime and lets us support more test use cases in a native image, it has the side effect of not providing the same key for the created `MergedContextConfiguration` as some of these implementations have an impact on the key.

We already know at build time which tests corresponds to which "context" as given by the created mapping:

```java
public class TestContextBootstrapInitializer {
  public static Map<String, Supplier<SmartContextLoader>> getContextLoaders() {
    Map<String, Supplier<SmartContextLoader>> entries = new HashMap<>();
    entries.put("org.springframework.samples.petclinic.owner.PetControllerTests", () -> new SpringBootAotContextLoader(PetControllerTestsContextInitializer.class, WebApplicationType.REACTIVE, SpringBootTest.WebEnvironment.NONE));
    entries.put("org.springframework.samples.petclinic.PetclinicIntegrationTests", () -> new SpringBootAotContextLoader(PetclinicIntegrationTestsContextInitializer.class, WebApplicationType.SERVLET, SpringBootTest.WebEnvironment.MOCK));
    entries.put("org.springframework.samples.petclinic.owner.OwnerControllerTests", () -> new SpringBootAotContextLoader(OwnerControllerTestsContextInitializer.class, WebApplicationType.REACTIVE, SpringBootTest.WebEnvironment.NONE));
    entries.put("org.springframework.samples.petclinic.vet.VetControllerTests", () -> new SpringBootAotContextLoader(VetControllerTestsContextInitializer.class, WebApplicationType.REACTIVE, SpringBootTest.WebEnvironment.NONE));
    entries.put("org.springframework.samples.petclinic.owner.VisitControllerTests", () -> new SpringBootAotContextLoader(VisitControllerTestsContextInitializer.class, WebApplicationType.REACTIVE, SpringBootTest.WebEnvironment.NONE));
    return entries;
  }
}
```

This mapping shows that all tests have an isolated context. On the other hand, the following mapping shows that the two tests share the same context:

```java
public class TestContextBootstrapInitializer {
  public static Map<String, Supplier<SmartContextLoader>> getContextLoaders() {
    Map<String, Supplier<SmartContextLoader>> entries = new HashMap<>();
[...]

PR #1341, opened 2021-12-05
Fix MergedContextConfiguration cache key for runtime contexts

sbrannen:
This is a fix for #1262.

A more detailed commit message will be crafted after peer review.


sbrannen:
Web support is not working with the status quo, so I'll have to take another look.

```
Caused by: java.lang.ClassCastException: class org.springframework.aot.test.AotMergedContextConfiguration cannot be cast to class org.springframework.test.context.web.WebMergedContextConfiguration (org.springframework.aot.test.AotMergedContextConfiguration and org.springframework.test.context.web.WebMergedContextConfiguration are in unnamed module of loader 'app')
	at org.springframework.aot.test.boot.AotSpringBootConfigContextLoader$WebConfigurer.configure(AotSpringBootConfigContextLoader.java:135)
	at org.springframework.aot.test.boot.AotSpringBootConfigContextLoader.loadContext(AotSpringBootConfigContextLoader.java:103)
	at org.springframework.aot.test.boot.AotSpringBootConfigContextLoader.loadContext(AotSpringBootConfigContextLoader.java:50)
	at org.springframework.aot.test.AotCacheAwareContextLoaderDelegate.loadContextInternal(AotCacheAwareContextLoaderDelegate.java:62)
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:124)
	... 68 common frames omitted
```

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
```diff
@@ -110,7 +116,7 @@ protected ClassName generateTestContext(BootstrapWriterContext writerContext, Su
 		return mainBootstrapClass.getClassName();
 	}
 
-	private MethodSpec.Builder contextLoadersMappingMethod(Map<ClassName, TestContextConfigurationDescriptor> entries) {
+	private MethodSpec.Builder getContextLoadersBuilder(Map<ClassName, TestContextConfigurationDescriptor> entries) {
```
snicoll, line 119: I am using such a pattern for all `MethodSpec.Builder` within the code base and not using the term "builder" in the name of the method was done on purpose as this is a private detail. The same goes for `get` that was removed. 

We can discuss any style change but this has to be done consistently and not as part of this PR.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
```diff
@@ -126,6 +132,34 @@ protected ClassName generateTestContext(BootstrapWriterContext writerContext, Su
 				.addModifiers(Modifier.PUBLIC, Modifier.STATIC).addCode(code.build());
 	}
 
+	private MethodSpec.Builder getContextInitializersBuilder(Map<ClassName, TestContextConfigurationDescriptor> entries) {
+		// We're generating a method that looks like the following.
```
snicoll, line 136: Please remove the comments. We now have a method that starts with an example while the existing one does not.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
```diff
@@ -126,6 +132,34 @@ protected ClassName generateTestContext(BootstrapWriterContext writerContext, Su
...
+		//     map.put("org.example.Sample2Tests", Sample2TestsContextInitializer.class);
+		//     return map;
+		// }
+
+		ClassName stringTypeName = ClassName.get(String.class);
+		TypeName initializerWildcard = WildcardTypeName.subtypeOf(Object.class);
+		TypeName initializerTypeName = ParameterizedTypeName.get(ClassName.get(ApplicationContextInitializer.class), initializerWildcard);
+		TypeName classWildcard = WildcardTypeName.subtypeOf(initializerTypeName);
+		TypeName classTypeName = ParameterizedTypeName.get(ClassName.get(Class.class), classWildcard);
+		TypeName mapTypeName = ParameterizedTypeName.get(ClassName.get(Map.class), stringTypeName, classTypeName);
```
snicoll, line 150: That's a lot of code to get `Map<String, Class<? extends ApplicationContextInitializer<?>>>`.  The following single statement does the same thing: 

```java
TypeName mapType = ParameterizedTypeName.get(ClassName.get(Map.class),
				ClassName.get(String.class), ParameterizedTypeName.get(ClassName.get(Class.class), WildcardTypeName.subtypeOf(ApplicationContextInitializer.class)));
```

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
```diff
@@ -126,6 +132,34 @@ protected ClassName generateTestContext(BootstrapWriterContext writerContext, Su
...
+
+		Builder code = CodeBlock.builder();
+		code.addStatement("$T map = new $T<>()", mapTypeName, HashMap.class);
```
snicoll, line 153: Detail but the field is called `entries` in the other method, why use something else here?

spring-aot-test/src/test/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessorTests.java
```diff
@@ -97,22 +98,41 @@ void processWritesContextLoaderMappingAtStandardLocation() {
...
 		ContextBootstrapStructure structure = this.tester.process(SampleApplicationTests.class);
 		assertThat(structure).hasClassDescriptor("com.example.TestContextBootstrapInitializer", (descriptor) -> {
-			assertThat(descriptor.getMethods()).singleElement().satisfies((methodDescriptor) -> {
-				assertThat(methodDescriptor.getName()).isEqualTo("getContextLoaders");
-				assertThat(methodDescriptor.getParameterTypes()).isEmpty();
-			});
 			assertThat(descriptor.getAccess()).isNull();
 			assertThat(descriptor.getFields()).isNull();
+			assertThat(descriptor.getMethods()).hasSize(2);
+			assertThat(descriptor.getMethods()).allSatisfy((methodDescriptor) -> {
```
snicoll, line 130: I am not fan of the | assertion here. We usually use `anySatisfy` when there is more than one element (there are examples in the codebase).

spring-aot-test/src/test/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessorTests.java
```diff
@@ -97,22 +98,41 @@ void processWritesContextLoaderMappingAtStandardLocation() {
...
+
 	@Test
 	void processInvokeTestNativeConfigurationRegistrar() {
 		ContextBootstrapStructure structure = this.tester.process(SampleApplicationTests.class);
 		assertThat(structure).hasResourcePattern("org/springframework/aot/test/samples/app/SampleApplication.class");
 	}
 
 	@Test
-	void processRegisterReflectionForContextLoadersMappingMethod() {
+	void processReflectionRegistrationForMappingMethods() {
```
snicoll, line 124: Please refrain from doing unrelated changes like this.

spring-native/src/main/java/org/springframework/aot/test/AotCacheAwareContextLoaderDelegate.java
```diff
@@ -38,22 +42,69 @@
...
+		this.aotContextLoader = aotContextLoader;
 	}
 
 	@Override
 	protected ApplicationContext loadContextInternal(MergedContextConfiguration config) throws Exception {
 		SmartContextLoader contextLoader = this.aotContextLoader.getContextLoader(config.getTestClass());
 		if (contextLoader != null) {
-			logger.info("Starting test in AOT mode using " + contextLoader);
-			return contextLoader.loadContext(config);
+			Assert.isInstanceOf(AotMergedContextConfiguration.class, config);
```
snicoll, line 62: These two lines are unrelated to the task here, isn't it?

spring-native/src/main/java/org/springframework/aot/test/AotCacheAwareContextLoaderDelegate.java
```diff
@@ -38,22 +42,69 @@
 
 	private final AotContextLoader aotContextLoader;
 
+	public AotCacheAwareContextLoaderDelegate() {
+		this(new AotContextLoader());
+	}
+
 	AotCacheAwareContextLoaderDelegate(AotContextLoader aotContextLoader) {
```
snicoll, line 49: This constructor is no longer called by anything as far as I can see, besides the default constructor. It should be removed.

spring-native/src/main/java/org/springframework/aot/test/AotContextLoader.java
```diff
@@ -70,6 +77,24 @@
 		}
 	}
 
+	@SuppressWarnings("unchecked")
```
snicoll, line 80: Isn't there a case to refactor the existing method to pass the method name and expected type? The cast is happening anyway so I don't see a lot of benefits in having two methods for the reflection business.

spring-native/src/main/java/org/springframework/aot/test/AotContextLoader.java
```diff
@@ -79,4 +104,8 @@ boolean isSupportedTestClass(Class<?> testClass) {
 		return this.contextLoaders.containsKey(testClass.getName());
 	}
 
+	Class<? extends ApplicationContextInitializer<?>> getContextInitializerClass(Class<?> testClass) {
```
snicoll, line 107: I am not against reusing the same component for both purposes (I actually think it's the right call considering the reflection business going on) but I think the class-javadoc should be updated and probably the class renamed as a result.

spring-native/src/main/java/org/springframework/aot/test/AotMergedContextConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+
+	private final MergedContextConfiguration original;
+
+
+	AotMergedContextConfiguration(Class<?> testClass,
+			Class<? extends ApplicationContextInitializer<?>> contextInitializerClass,
+			MergedContextConfiguration original, AotCacheAwareContextLoaderDelegate cacheAwareContextLoaderDelegate,
+			@Nullable MergedContextConfiguration parent) {
+
+		super(testClass, null, null, Collections.singleton(contextInitializerClass), null, null,
```
snicoll, line 54: `ContextLoader` can't be null according to the contract. We can ignore that if we choose to but I thought it was worth raising.

spring-native/src/test/java/org/springframework/aot/test/AotCacheAwareContextLoaderDelegateTests.java
```diff
@@ -45,33 +52,107 @@
...
-		assertThat(delegate.loadContextInternal(createMergedContextConfiguration(SampleTest.class)))
-				.isSameAs(this.applicationContext);
-		verify(this.contextLoader).loadContext(any(MergedContextConfiguration.class));
+	void loadContextWithMatchUsesAotInfrastructure() throws Exception {
+		AotCacheAwareContextLoaderDelegate delegate = createDemoDelegate();
+
+		MergedContextConfiguration mergedConfig = createMergedContextConfiguration(SampleTest.class);
+		assertThat(delegate.loadContext(mergedConfig)).isSameAs(this.applicationContext);
+
+		// loadContext() must actually be invoked with the original MergedContextConfiguration
```
snicoll, line 69: It will take me a bit of time to process this but, at a quick glance, I generally feel that a "lot' of explanations like that in a test is the sign that something else should be made more explicit.

spring-native/src/main/java/org/springframework/aot/test/AotMergedContextConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+	/**
+	 * Get the original {@link MergedContextConfiguration} that this
+	 * {@code AotMergedContextConfiguration} was created for.
+	 */
+	MergedContextConfiguration getOriginal() {
+		return this.original;
+	}
+
+	@Override
+	public boolean equals(@Nullable Object other) {
```
snicoll, line 69: I see 0 test hits on this method which seems a little strange.

snicoll (commented):
As requested by @sdeleuze, here's a first review. Happy to chat if more details are needed.

sdeleuze:
Please let me know when this PR is ready for final review and merge.

sbrannen:
> Please let me know when this PR is ready for final review and merge.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
sbrannen: OK. I've reverted that and renamed the new method to `contextInitializersMappingMethod` for consistency.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
sbrannen: I disagree with the rationale.

When JavaPoet is used to generate code, it is not readily clear what is being generated. Thus, the inclusion of the comment is beneficial to anyone reading the code and especially to anyone who has to modify or maintain the code.

Instead of removing this comment, I suggest we add a similar comment to `contextLoadersMappingMethod()` for consistency.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
sbrannen: The existing code produces:

```java
Map<String, Class<? extends ApplicationContextInitializer<?>>>
```

The proposed change produces the following which loses the `<?>`:

```java
Map<String, Class<? extends ApplicationContextInitializer>>
```

Although the proposed change is technically a single statement, it's 202 characters long, and I don't think it improves readability or maintainability of the code.

In any case, I'll see about reducing the number of local variables currently in use.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
sbrannen: Speaking of "field", I actually was wondering about converting it from a local variable to a `static` field, since we're currently recreating the maps each time one of these methods is invoked (which happens a lot for `getContextInitializers()`).

Had you considered that?

Was there a technical reason to create the map as a local variable instead of as a static field?

sbrannen: With regard to the name of the variable, every map contains entries. So unless the term "entries" has special meaning for the context in which a map is used, I always prefer the generic term `map` unless I think it's necessary to name the map something  more specific.

Since this is (currently) a local variable, I don't see the need for anything more specific than `map`.

With regard to using something different here, I didn't want to change the implementation of the other method, because I know you do not like changes that you do not consider relevant to the current PR.

spring-aot-test/src/test/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessorTests.java
sbrannen: OK. I'll take a look at other usage of `anySatisfy` and see if I can apply that here.

spring-aot-test/src/test/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessorTests.java
sbrannen: Handled separately in 4787862c319fc609821922102f41e991abb5b1ef

spring-native/src/main/java/org/springframework/aot/test/AotCacheAwareContextLoaderDelegate.java
sbrannen: The log statement has been handled separately in 4787862c319fc609821922102f41e991abb5b1ef.

The other line is related to this PR.

spring-native/src/main/java/org/springframework/aot/test/AotCacheAwareContextLoaderDelegate.java
sbrannen: Good catch! 👍 

I simply overlooked that.

It's gone now.

spring-native/src/main/java/org/springframework/aot/test/AotMergedContextConfiguration.java
sbrannen: Yeah, I noticed that as well, but thanks for noticing and mentioning it (b/c I might have overlooked it)!

A non-null `ContextLoader` is in fact required by `DefaultCacheAwareContextLoaderDelegate.loadContextInternal(...)`, but we override that method and ignore the `ContextLoader` in the supplied `MergedContextConfiguration` since we are using the `AotSpringBootConfigContextLoader`.

So I think it's safe to ignore this for our use case.

spring-native/src/main/java/org/springframework/aot/test/AotMergedContextConfiguration.java
sbrannen: `equals()` and `hashCode()` are used by the `LruCache` in `DefaultContextCache`.

So if you don't see any hits, that means that the context cache is not being exercised.

spring-native/src/test/java/org/springframework/aot/test/AotCacheAwareContextLoaderDelegateTests.java
sbrannen: Fair enough.

I'll move the comments to actual Javadoc in the production code.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
snicoll: Thanks but I still don't think this is a good idea, certainly for the reason you've shared. 

The tests are meant to show what is generated rather than a comment that is going to be outdated. If you need a visual representation of what the method will look like concretely, write a test for it that asserts the content. That's what do everywhere else.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
snicoll: Not going to comment any further on the 202 remark besides the fact that loosing variable assignment was the goal here. Sorry I missed the extra wildcard.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
snicoll: > With regard to using something different here, I didn't want to change the implementation of the other method, because I know you do not like changes that you do not consider relevant to the current PR.

That's exactly the point that I am trying to raise for several PRs in a row now. If you disagree about something, that's fine to discuss it and then come to a conclusion but changing it in your own code just brings inconsistency and I don't like that.

spring-native/src/main/java/org/springframework/aot/test/AotMergedContextConfiguration.java
snicoll: > So if you don't see any hits, that means that the context cache is not being exercised.

Exactly. The context cache is not exercised while running all tests in that package concerns me. Shouldn't it be used?

spring-native/src/main/java/org/springframework/aot/test/AotContextLoader.java
sbrannen: I was actually considering doing that but hadn't gotten to it.

It's done now (locally).

spring-native/src/main/java/org/springframework/aot/test/AotContextLoader.java
sbrannen: I agree. 

I'll come up with a new name and update the documentation.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
sbrannen: OK. I'll remove the comment.

I already have a test for it that asserts the content.

spring-aot-test/src/main/java/org/springframework/aot/test/context/bootstrap/generator/TestContextAotProcessor.java
sbrannen: > Speaking of "field", I actually was wondering about converting it from a local variable to a static field, since we're currently recreating the maps each time one of these methods is invoked (which happens a lot for getContextInitializers()).

You can disregard that. I think that used to be the case, but it no longer is. The methods that create the maps should only be invoked twice now: once for `AotCacheAwareContextLoaderDelegate` and once for `AotDependencyInjectionTestExecutionListener`.

spring-native/src/main/java/org/springframework/aot/test/AotMergedContextConfiguration.java
sbrannen: Ahhh... I mistakenly thought you were questioning the need for overriding `equals()`.

> Shouldn't it be used?

It's used in the sample projects as functional tests, but I'll see if I can introduce an integration test that exercises those methods via the context cache.

snicoll: > Ahhh... I mistakenly thought you were questioning the need for overriding equals().

On the contrary. Implementing `equals` and `hashCode` is at the hear of the PR so not having tests for those felt strange. Simple unit tests should be enough though, no need for having to deal with the context cache.

sbrannen:
This PR is ready for a final review.

The only remaining task is an update to the Javadoc (see https://github.com/spring-projects-experimental/spring-native/pull/1341#discussion_r763187815), but that can be done at later date if necessary.

snicoll:
Thanks Sam, this looks good.

PR #1346, opened 2021-12-06
Avoid creating reflection entries for java and primitive types.

christophstrobl:
This commit fixes an issue where the JPA processor inspects too many types for potential reflection configuration. We now skip types within the `java` namespace as well as primitives and primitive arrays.


sdeleuze:
@schauder I am hesitant about what is the right behavior for JPA native reflection configuration for Java/primitive types. Currently it provide full reflection for all methods/constructors/fields of java.* types, that's too much for sure. But should we skip those totally or just configure reflection on related class without methods/constructors/fields flags (depending on what JPA/Hibernate require)?

PR #1347, opened 2021-12-06
Register reflection configuration for JPA AttributeConverter implementations

christophstrobl:
This commit makes sure to look up `@Converter` annotated `AttributeConverter` implementations within the base packages and adds required reflection configuration for both the type itself as well as the conversion return type.

Depends on (includes): #1346


sdeleuze:
Merged via 906925e9d4e0b5c13fe755b86d934dc7eb479aa8, thanks.

PR #1351, opened 2021-12-06
Provide implementation of BeanFactoryNativeConfigurationProcessor for function types

olegz:
Add initial implementation of FunctionTypeProcessor and FunctionTypeProcessorTests
Modified samples to remove TypeHint as well as reference to it in the README
Upgrade Spring Cloud dependency to 2021.0.0


spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+			new BeanFactoryProcessor(beanFactory).processBeans(this::isFunction,
+				(beanName, functionBeanType) -> {
+					Type functionType = FunctionTypeUtils.discoverFunctionTypeFromClass(functionBeanType);
+					Type inputType = FunctionTypeUtils.getInputType(functionType);
+					String name = inputType.getTypeName();
+					if (!name.startsWith("java.") &&
+						!name.startsWith("javax.")) {
+						if (added.add(name)) {
+							registry.reflection().forType(FunctionTypeUtils.getRawType(inputType))
+								.withAccess(TypeAccess.DECLARED_CONSTRUCTORS, TypeAccess.PUBLIC_CONSTRUCTORS, TypeAccess.DECLARED_FIELDS, TypeAccess.PUBLIC_FIELDS, TypeAccess.DECLARED_METHODS, TypeAccess.PUBLIC_METHODS);
```
sdeleuze, line 59: Are `TypeAccess.DECLARED_FIELDS` and `TypeAccess.PUBLIC_FIELDS` really needed here?

spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+						}
+					}
+				});
+		}
+
+		private boolean isFunction(Class<?> beanType) {
+			return Function.class.isAssignableFrom(beanType)
+					|| Consumer.class.isAssignableFrom(beanType);
```
sdeleuze, line 67: Could we check via the presence of a bean type in `beanFactory` that Spring Cloud Function is enabled in this application?

sdeleuze (changes requested):

spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
olegz: Well, the fact that I am using `FunctionTypeUtils` means s-c-function is on the classpath. We don't have any explicit _enable_ functionality, hence my earlier question, how do these processors work when the "optional" dependency they are using is not available.  So I am not sure what else i can add here

spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
olegz: Probably not. It was mainly cut/paste from the 'alike' component, so i'll fix it

samples/cloud-function-netty-fun/result.txt
```diff
@@ -0,0 +1 @@
+
```
sdeleuze, line 1: Should be removed right?

spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
snicoll: @olegz you got the `Processor` nested class pattern right, except you're not checking any condition so that makes it useless. See https://github.com/spring-projects-experimental/spring-native/blob/906925e9d4e0b5c13fe755b86d934dc7eb479aa8/spring-aot/src/main/java/org/springframework/boot/actuate/endpoint/annotation/EndpointNativeConfigurationProcessor.java#L49 for example.


> Well, the fact that I am using FunctionTypeUtils means s-c-function is on the classpath.

Right now if you run this hint without Spring Cloud Function, it will fail with NCDFE.

olegz: Yeah, sorry missed it. That is what I was asking Sebastien earlier about `Conditional`. Anyway, let me fix that

spring-aot/src/main/resources/META-INF/spring.factories
```diff
@@ -31,6 +31,7 @@ org.springframework.web.WebNativeConfigurationProcessor,\
 org.springframework.core.SynthesizedAnnotationNativeConfigurationProcessor,\
 org.springframework.core.IndexedBeanHierarchyNativeConfigurationProcessor,\
 org.springframework.context.annotation.ScopeNativeConfigurationProcessor
+org.springframework.cloud.function.FunctionTypeProcessor
```
oclay1st, line 34: @olegz, @sdeleuze  I think there was a mistake after merging this PR, there is a missing comma after org.springframework.context.annotation.ScopeNativeConfigurationProcessor

PR #1362, opened 2021-12-09
Use AssertableApplicationContext in Spring Cloud sample tests

OlgaMaciaszek:


snicoll:
@OlgaMaciaszek thanks for following-up!

Issue #1368, opened 2021-12-11
spring-data-jpa @EnableJpaRepositories not supported?


liupengpop:
@EnableJpaRepositories(repositoryBaseClass = MyJpaRepository.class)
not supported?

***************************
APPLICATION FAILED TO START
***************************

Description:

Native reflection configuration for com.test.MyJpaRepository is missing.

Action:

Native configuration for a class accessed reflectively is likely missing.
You can try to configure native hints in order to specify it explicitly.
See https://docs.spring.io/spring-native/docs/current/reference/htmlsingle/#native-hints for more details.

snicoll:
@liupengpop it is hard to give you support with this level of details. Spring Data JPA is supported. You haven't shared which version of Spring Native you're using or what your setup looks like. For us to be able to help you, we'd need that information.

PR #1383, opened 2021-12-17
Register reflection configuration for custom Spring Data repository base class.

christophstrobl:
This PR makes sure to read a potentially given custom repository base class implementation and register reflection configuration for it.

Resolves: #1368

### DIFF
--- a/spring-aot/src/main/java/org/springframework/data/RepositoryDefinitionConfigurationProcessor.java
+++ b/spring-aot/src/main/java/org/springframework/data/RepositoryDefinitionConfigurationProcessor.java
@@ -122,6 +122,7 @@ public class RepositoryDefinitionConfigurationProcessor implements BeanFactoryNa
 		private static final String FRAGMENTS_PROPERTY = "repositoryFragments";
 
 		private static final String CUSTOM_IMPLEMENTATION_PROPERTY = "customImplementation";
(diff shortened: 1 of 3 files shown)


spring-aot/src/main/java/org/springframework/data/RepositoryDefinitionConfigurationProcessor.java
```diff
@@ -239,6 +259,16 @@ void writeConfiguration(RepositoryConfiguration configuration) {
...
+			writeCustomRepositoryBaseClass(configuration);
+		}
+
+		private void writeCustomRepositoryBaseClass(RepositoryConfiguration configuration) {
+
+			if (!configuration.hasCustomBaseClass()) {
+				return;
+			}
+
+			registry.reflection().forType(configuration.getCustomRepositoryBaseClass()).withAccess(TypeAccess.DECLARED_CONSTRUCTORS, TypeAccess.DECLARED_METHODS);
```
sdeleuze, line 271: What about adding `TypeAccess.PUBLIC_METHODS` in order to have also the inherited ones?

christophstrobl: just public ones should actually do.

Issue #1405, opened 2021-12-26
`data-jpa` sample native image won't start with `@EnableJpaRepositories`


ch4mpy:
If adding `@EnableJpaRepositories`, regular image (_non-native_) start correctly but native image won't start:
- `mvn clean spring-boot:run` => OK 
- `mvn clean spring-boot:build-image && docker run --rm -p 8080:8080 -t data-jpa:0.0.1-SNAPSHOT` => KO
```
2021-12-26 11:22:00.841  INFO 1 --- [           main] j.LocalContainerEntityManagerFactoryBean : Initialized JPA EntityManagerFactory for persistence unit 'default'
2021-12-26 11:22:00.858  WARN 1 --- [           main] w.s.c.ServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.spr
ingframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'app.main.SampleApplication': Unsatisfied dependency expressed through constructor parameter 0; n
ested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'fooRepository': Invocation of init method failed; nested exception is com.oracl
e.svm.core.jdk.UnsupportedFeatureError: Proxy class defined by interfaces [interface org.springframework.data.jpa.repository.support.CrudMethodMetadata, interface org.springframework.aop
.SpringProxy, interface org.springframework.aop.framework.Advised, interface org.springframework.core.DecoratingProxy] not found. Generating proxy classes at runtime is not supported. Pr
oxy classes need to be defined at image build time by specifying the list of interfaces that they implement. To define proxy classes use -H:DynamicProxyConfigurationFiles=<comma-separate
d-config-files> and -H:DynamicProxyConfigurationResources=<comma-separated-config-resources> options.
2021-12-26 11:22:00.858  INFO 1 --- [           main] j.LocalContainerEntityManagerFactoryBean : Closing JPA EntityManagerFactory for persistence unit 'default'
2021-12-26 11:22:00.858  INFO 1 --- [           main] .SchemaDropperImpl$DelayedDropActionImpl : HHH000477: Starting delayed evictData of schema as part of SessionFactory shut-down'
2021-12-26 11:22:00.858  INFO 1 --- [           main] com.zaxxer.hikari.HikariDataSource       : HikariPool-1 - Shutdown initiated...
[...]

PR #1406, opened 2021-12-26
Demo gh-1405

ch4mpy:
Demo for reproducing gh-1405: if spring-boot app is decorated with `@EnableJpaRepositories`, 
- "regular" images start correctly (`mvn clean spring-boot:run`)
- "native" image won't (`mvn clean spring-boot:build-image && docker run --rm -p 8080:8080  -t data-jpa:0.0.1-SNAPSHOT`) .


sdeleuze:
For for the repro, this will be fixed and checked in the sample via #1405.

Issue #769, opened 2021-05-06
Add Coroutines reflection inference


sdeleuze:
See https://github.com/spring-projects-experimental/spring-native/issues/643#issuecomment-833381037.

Can be implemented by using `kotlin-reflect` in Spring Native or via https://github.com/spring-projects/spring-framework/issues/21546.

PR #1432, opened 2022-01-09
make reflection analysis cover kotlin coroutines usage

czp3009:
If a controller method is kotlin coroutine suspend fun, it's return type will be `Continuation<? super Bar>`

In this situation, `returnType.getActualTypeArguments()[0]` will be `? super Bar`, it's a WildcardType, and lower bounds is the exact return type `Bar`.

So we need to dig out the type `Bar` from the original wrapped Continuation type.

If just skip the WildcardType, suspend fun will be considered as a method which return Object.

Adding a lot of @TypeHint is very bad for migrating old projects to spring native, I want to analyze directly in spring aot to get the actual return value of suspend fun.

### DIFF
--- a/samples/webflux-kotlin/src/main/kotlin/com/example/webflux/WebfluxApplication.kt
+++ b/samples/webflux-kotlin/src/main/kotlin/com/example/webflux/WebfluxApplication.kt
@@ -7,8 +7,6 @@ import org.springframework.nativex.hint.NativeHint
 import org.springframework.nativex.hint.TypeAccess
 import org.springframework.nativex.hint.TypeHint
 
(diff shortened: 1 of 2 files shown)


sdeleuze:
Interesting proposal that could maybe fix #769 more easily than originally expected, could you please update [webflux-kotlin](https://github.com/spring-projects-experimental/spring-native/blob/main/samples/webflux-kotlin/src/main/kotlin/com/example/webflux/WebfluxApplication.kt#L10-L11) to remove those manual hints and see if that works?

Also could you please update the PR to perform this type inferience only when this type is a `Continuation` to avoid side effects and keep an optimal footprint when that's not required?

sdeleuze:
Merged with additional refinements via #769, thanks a lot for this great PR!

PR #1442, opened 2022-01-12
Fix data repository type resolution.

christophstrobl:
This commit makes sure to capture all required generic types from the repository interface definition instead of just picking selected ones.


christophstrobl:
invalid - did not change a thing.

PR #1447, opened 2022-01-13
Support "args" param in "@SpringBootTest"

ttddyy:
Reflectively access `SpringBootTestArgs`(package private) to support `args` param in `@SpringBootTest`.

### DIFF
--- a/spring-native/src/main/java/org/springframework/aot/test/boot/AotSpringBootConfigContextLoader.java
+++ b/spring-native/src/main/java/org/springframework/aot/test/boot/AotSpringBootConfigContextLoader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2021 the original author or authors.
+ * Copyright 2019-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


spring-native/src/main/java/org/springframework/aot/test/boot/AotSpringBootConfigContextLoader.java
```diff
@@ -49,6 +55,23 @@
  */
 public class AotSpringBootConfigContextLoader extends SpringBootContextLoader {
 
+	// "SpringBootTestArgs#get(Set)" static method
+	private static final Method SPRING_BOOT_TEST_ARGS__GET;
```
snicoll, line 59: Rather than using reflection, I'd rather add an accessor class in `spring-native` the same way we do elsewhere (a class in the same package that exposes the package-private behavior).

snicoll (commented):
Thanks for the PR @ttddyy. This is an interesting approach but I'd rather compute the arguments upfront and pass them to the loader instead. This means changing the loader and the source code generator that uses it. We can add a vararg for now so that the code without specific arguments are the same.

ttddyy:
Thanks @snicoll, I got a picture. 
Yeah, since `MergedContextConfiguration` is available at aot build time, it should pre-calculate the args. I'll update the change.

sdeleuze:
@snicoll Should we include that in 0.11.2? If yes, can I let you merge it with potential refinement if needed?

snicoll:
Thanks for the nudge. It looks good but considering we're going to release shortly, I've moved that one to the next release.

snicoll:
Thanks again @ttddyy

PR #1460, opened 2022-01-20
"@EnabledOnAot" and "@EnabledOnNative" annotations

ttddyy:
I'm experimenting `@EnabledOnAot`, `@DisabledOnAot`, `@EnabledOnNative`, and `@DisabledOnNative` annotations.

These are directly implementing JUnit5 `ExecutionCondition`.
Another approach would be to meta annotate with `@EnabledIf` with SpEL. But it requires a test to be a spring test. So, instead, I am making them direct Junit5 extensions.

### DIFF
--- a/spring-native-docs/src/main/asciidoc/getting-started-native-build-tools.adoc
+++ b/spring-native-docs/src/main/asciidoc/getting-started-native-build-tools.adoc
@@ -369,3 +369,10 @@ $ gradle nativeTest
 ----
 
 You can find more details about the native build tools https://github.com/graalvm/native-build-tools[here].
+
+To conditionally enable/disable tests, we provide the following annotations:
+
+- `@EnabledOnAot`
+- `@DisabledOnAot`
+- `@EnabledOnNative`
+- `@DisabledOnNative`
--- a/spring-native/pom.xml
+++ b/spring-native/pom.xml
@@ -62,6 +62,11 @@
 			<artifactId>spring-boot-test</artifactId>
 			<scope>provided</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-api</artifactId>
+			<scope>provided</scope>
+		</dependency>
 
 		<dependency>
 			<groupId>jakarta.validation</groupId>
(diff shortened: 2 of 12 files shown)


sdeleuze:
@snicoll @bclozel @sbrannen Any take of if we should ship this is as part of Spring Native `0.11.3` to get feedback from users and experiment ourselves on that?

snicoll:
I'd be more comfortable if we had a story of our own about this and, unless I've missed it, we don't. I know that we want to run some tests and make sure they work in Native once the necessary hints have been provided. Perhaps we should do that first and check if adding such annotations makes writing tests more idiomatic.

I don't really see a need for the `@EnabledOnAot`.

sdeleuze:
I think the need for `@EnabledOnAot` mostly comes from current limitation like lack of Mockito support, but that will hopefully not be the case in Spring Boot 3 anymore, so I tend to agree about lack of significant use case middle term.

For `@EnableOnNative`, there is likely more use cases, but I am wondering if this coul be implemented at JUnit5 level in addition to the others [Java runtime environment conditions](https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-jre) using the [`org.graalvm.nativeimage.imagecode` system property](https://github.com/oracle/graal/blob/master/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/ImageInfo.java). @sbrannen Any thoughts?

sbrannen:
> For `@EnableOnNative`, there is likely more use cases, but I am wondering if this could be implemented at JUnit5 level in addition to the others [Java runtime environment conditions](https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-jre) using the [`org.graalvm.nativeimage.imagecode` system property](https://github.com/oracle/graal/blob/master/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/ImageInfo.java). @sbrannen Any thoughts?

That is certainly imaginable. Though I don't think it would fit with the existing `@EnabledOnJre`. Rather, I think it would require its own dedicated `ExecutionCondition` and associated annotation.

If you'd like to see if the JUnit team is up for that, feel free to [open an issue](https://github.com/junit-team/junit5/issues/new/choose), and we'll discuss it.

sdeleuze:
Ok I have created https://github.com/junit-team/junit5/issues/2830 since I think that's a real use case but likely not something we want to implement at Spring Native level.

@ttddyy As a consequence,  I am closing this PR unmerged but thanks for your contribution which allowed to refine that use case and let see how it goes on JUnit side.

Issue #1463, opened 2022-01-23
Classpath for gradle built jar not the same as maven jar, aot.jar produced by gradle plugin is last in classpath


hdeadman:
After this [refresh scope](https://github.com/spring-projects-experimental/spring-native/issues/1413) issue was fixed, I was able to get passed the error with a native build but if I tried to build and run the aot enabled jar in java, it didn't work, but it worked in a maven reproducer project. I believe the issue has to do with the classpath in the gradle produced jar. I think the maven plugin puts aot files in BOOT-INF/classes (before the jars) and the gradle plugin adds a jar `projectname-aot.jar` but it is listed last in the classpath.idx file. 

I have a reproducer project that can be built with gradle and maven here: https://github.com/hdeadman/spring-native-reproducer

You can see the maven build and gradle build in the [Github Actions workflow](https://github.com/hdeadman/spring-native-reproducer/actions/runs/1734808311) or you can run `./test-maven.sh` and `./test-gradle.sh` to see the problem locally. The gradle built jar doesn't start up b/c some AOT code doesn't override the regular code. 

A similar problem was fixed for tests by this issue: https://github.com/spring-projects-experimental/spring-native/issues/1272 fixed by https://github.com/spring-projects-experimental/spring-native/commit/eba6bc43b53ee3fb07f9378064f67af4641ea556

bclozel:
We might need to prepend the generated Jar to the class path of the bootJar task [here](https://github.com/spring-projects-experimental/spring-native/blob/main/spring-aot-gradle-plugin/src/main/java/org/springframework/aot/gradle/SpringAotGradlePlugin.java#L138).

bclozel:
Superseded by #1466

PR #1466, opened 2022-01-23
Prepend AOT jar to classpath for bootJar Gradle task

hdeadman:
Have the spring aot gradle plugin put generated aot jar at the beginning of the boot jar classpath rather than the end. 

Fixes https://github.com/spring-projects-experimental/spring-native/issues/1463


bclozel:
Thanks for your contribution @hdeadman !

Issue #1473, opened 2022-01-28
ClassNotFoundException: antlr.CommonToken


ahs441:
I have an issue when running native spring boot app on Gradle with nativeRun or bootBuildImage and then run on docker;
when the application is launching after checking JPA the following exception accrue and application is failed to launch
CharScanner; panic: ClassNotFoundException: antlr.CommonToken

pluings I used:
plugins {
    id 'org.springframework.boot' version '2.6.3'
    id 'io.spring.dependency-management' version '1.0.11.RELEASE'
    id 'java'
    id 'org.springframework.experimental.aot' version '0.11.2'
    id 'org.hibernate.orm'
}

build image config:
bootBuildImage {
    builder = 'paketobuildpacks/builder:tiny'
    environment = ['BP_NATIVE_IMAGE': 'true']
    imageName='ahs441/vusicnew'
}

hibernate config:
hibernate {
    enhance {
        enableLazyInitialization = true
        enableDirtyTracking = true
        enableAssociationManagement = true
        enableExtendedEnhancement = false
    }
}

bclozel:
We're going to need more information to understand what's going on here. A complete stacktrace would help, but I suspect a minimal, sample application would be more helpful. Could you share a minimal repro app for this?

ahs441:
The stacktrace is empty with only CharScanner; panic: ClassNotFoundException: antlr.CommonToken exception;
The dependencies are:
 implementation 'org.springframework.boot:spring-boot-starter-actuator'
    implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
    implementation 'org.springframework.boot:spring-boot-starter-security'
    implementation 'org.springframework.boot:spring-boot-starter-validation'
    implementation 'org.springframework.boot:spring-boot-starter-web'
    annotationProcessor "org.springframework.boot:spring-boot-configuration-processor"
    compileOnly 'org.projectlombok:lombok'
    runtimeOnly 'org.postgresql:postgresql'
    annotationProcessor 'org.projectlombok:lombok'
    testImplementation 'org.springframework.boot:spring-boot-starter-test'
    testImplementation 'org.springframework.security:spring-security-test'

    implementation 'io.jsonwebtoken:jjwt-api:0.11.2'
    runtimeOnly 'io.jsonwebtoken:jjwt-impl:0.11.2',
            'io.jsonwebtoken:jjwt-jackson:0.11.2' 

[...]

PR #1476, opened 2022-02-01
Add missing hints to support maintaining order in an element collection

christophstrobl:
Provide required hints to support maintaining order in element collection jpa/hibernate.

```java
@OrderBy
@ElementCollection
private List<String> phoneNumbers;
```

----

To avoid static hints in future versions we could consider moving the hints into `JpaConfigurationProcessor` only registering them when a trigger annotation is used within the domain model.

Resolves: #1473


sdeleuze:
Thanks, should probably be moved to the native configuration repository once available.

PR #1496, opened 2022-02-15
Add reflective registration of RootBeanDefinition to FunctionTypeProcessor

olegz:
- polish dependencies in cloud-function-aws POM

### DIFF
--- a/spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
+++ b/spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.function;
 
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
```diff
@@ -60,6 +61,17 @@ void process(ConfigurableListableBeanFactory beanFactory, NativeConfigurationReg
 											TypeAccess.DECLARED_METHODS, TypeAccess.PUBLIC_METHODS);
 						}
 					});
+			registry.reflection().forType(RootBeanDefinition.class).withAccess(TypeAccess.DECLARED_FIELDS);
```
sdeleuze, line 64: Could we just add the required field instead of all of them?
Also if that's static, maybe that should go to `FunctionHints` instead of here.

spring-aot/src/main/java/org/springframework/cloud/function/FunctionTypeProcessor.java
```diff
@@ -60,6 +61,17 @@ void process(ConfigurableListableBeanFactory beanFactory, NativeConfigurationReg
...
+			try {
```
sdeleuze, line 65: Same remark than above, maybe better in `FunctionHints`.

spring-aot/src/test/java/org/springframework/cloud/function/FunctionTypeProcessorTests.java
```diff
@@ -39,20 +40,35 @@
  */
 class FunctionTypeProcessorTests {
 
-	private static final Set<TypeAccess> ALL_MEMBERS = Set.of(TypeAccess.DECLARED_CONSTRUCTORS,
+	private static final Set<TypeAccess> METHODS_AND_CONSTRUCTORS = Set.of(TypeAccess.DECLARED_CONSTRUCTORS,
```
sdeleuze, line 43: Instead of testing here, could we just add an `@Bean` use case and modify `verify.sh` in one of the `cloud-function-*` sample?

sdeleuze (changes requested):

spring-aot/src/test/java/org/springframework/cloud/function/FunctionTypeProcessorTests.java
olegz: Let me try that

olegz:
Will submit a new PR as I need to clean up few things

Issue #1514, opened 2022-03-01
Spring Aot Maven Plugin does not work with Maven Compiler Plugin 3.9.0


snicoll:
Using `maven-compiler-plugin` 3.9.0 or later leads to the following exception when the AOT plugin tries to instrument it:

```
Caused by: org.apache.maven.plugin.PluginConfigurationException: Unable to parse configuration of mojo org.apache.maven.plugins:maven-compiler-plugin:3.9.0:testCompile for parameter compilePath: Cannot find 'compilePath' in class org.apache.maven.plugin.compiler.TestCompilerMojo
    at org.apache.maven.plugin.internal.DefaultMavenPluginManager.populatePluginFields (DefaultMavenPluginManager.java:665)
    at org.apache.maven.plugin.internal.DefaultMavenPluginManager.getConfiguredMojo (DefaultMavenPluginManager.java:597)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:124)
    at org.twdata.maven.mojoexecutor.MojoExecutor.executeMojo (MojoExecutor.java:120)
    at org.springframework.aot.maven.TestGenerateMojo.compileGeneratedTestSources (TestGenerateMojo.java:178)
    at org.springframework.aot.maven.TestGenerateMojo.execute (TestGenerateMojo.java:154)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:210)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
```

mhalbritter:
[...]

PR #1559, opened 2022-03-30
Spring Aot Maven Plugin does not work with Maven Compiler Plugin > 3.8.1

mhalbritter:
- Updates the maven-compiler-plugin in the samples to the latest available one, 3.10.1

- The maven-compiler-plugin starting from 3.9.0 removed the configuration option 'compilePath', which the TestGenerateMojo was using. This option was not used and have been removed by the maven-compiler-plugin team in commit https://github.com/apache/maven-compiler-plugin/commit/419bf040893dac8f300deb5b36daf5f487dea8d2

See gh-1514

### DIFF
--- a/spring-aot-maven-plugin/src/main/java/org/springframework/aot/maven/TestGenerateMojo.java
+++ b/spring-aot-maven-plugin/src/main/java/org/springframework/aot/maven/TestGenerateMojo.java
@@ -38,7 +38,6 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.apache.maven.shared.utils.cli.CommandLineUtils;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
-import org.twdata.maven.mojoexecutor.MojoExecutor;
 
 import org.springframework.aot.test.build.GenerateTestBootstrapCommand;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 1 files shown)


sdeleuze:
Thanks for this PR, does the AOT plugin still work with previous versions (<= 3.8.1) or not?

@bclozel Could you please review the PR?

mhalbritter:
Oh sorry, should have put that in the description. Yes, i tested it with 3.8.1 too, and it works. The `compilePath` is a noop in 3.8.1.

samples/maven-parent/pom.xml
```diff
@@ -28,11 +28,17 @@
 		<wavefront.version>2.2.0</wavefront.version>
 		<spring-cloud.version>2021.0.1</spring-cloud.version>
 		<native.buildtools.version>0.9.10</native.buildtools.version>
+		<maven-compiler-plugin.version>3.10.1</maven-compiler-plugin.version>
```
snicoll, line 31: I would not do that. This is managing by Spring Boot already so we should use whatever it provides.

samples/maven-parent/pom.xml
```diff
@@ -28,11 +28,17 @@
...
 	</properties>
 
 	<build>
 		<pluginManagement>
 			<plugins>
+				<plugin>
```
snicoll, line 37: That's not necessary as the parent already manages the plugin.

snicoll (commented):
Thanks for the PR. I've added a suggestion, let me know what you think.

samples/maven-parent/pom.xml
mhalbritter: Yes, you're absolutely right. I didn't realize that the starter parents inherit from the spring-boot-starter. I will remove the `pluginManagement` and the version override property.

Issue #1561, opened 2022-03-30
Native-tests are broken with GraalVM 22.1


mhalbritter:
Running for example the tests in `commandlinerunner` via `mvn test -Pnative`

## Working version:

```
openjdk version "17.0.2" 2022-01-18
OpenJDK Runtime Environment GraalVM CE 22.0.0.2 (build 17.0.2+8-jvmci-22.0-b05)
OpenJDK 64-Bit Server VM GraalVM CE 22.0.0.2 (build 17.0.2+8-jvmci-22.0-b05, mixed mode, sharing)
```

[Working output](https://github.com/spring-projects-experimental/spring-native/files/8380971/22.0.txt)

## Broken version (built from https://github.com/oracle/graal/commit/3eb1d111ab25565acbbc015128a2c787656e2675):

```
openjdk version "17.0.3" 2022-04-19
OpenJDK Runtime Environment GraalVM 22.1.0-dev (build 17.0.3+4-jvmci-22.1-b02)
OpenJDK 64-Bit Server VM GraalVM 22.1.0-dev (build 17.0.3+4-jvmci-22.1-b02, mixed mode, sharing)
```

[Broken output](https://github.com/spring-projects-experimental/spring-native/files/8380984/22.1.txt)

Main problem seems to be

```
com.oracle.svm.core.jdk.UnsupportedFeatureError: Proxy class defined by interfaces [interface org.mockito.plugins.PluginSwitch] not found. Generating proxy classes at runtime is not supported. Proxy classes need to be defined at image build time by specifying the list of interfaces that they implement. To define proxy classes use -H:DynamicProxyConfigurationFiles=<comma-separated-config-files> and -H:DynamicProxyConfigurationResources=<comma-separated-config-resources> options.
```

sdeleuze:
Could you try to see if we can exclude it via `MockitoExcludeFilter`?

mhalbritter:
https://github.com/spring-projects-experimental/spring-native/pull/1562 fixes this issue by using `IgnoredFactoriesCodeContributor` to ignore `MockitoTestExecutionListener` and `ResetMocksTestExecutionListener` when generating AOT code.

PR #1562, opened 2022-03-31
Fix native test failures with Graal 22.1

mhalbritter:
- The failures occur because MockitoTestExecutionListener and
  ResetMocksTestExecutionListener tried to initialize Mockito,
  which isn't supported yet.
- This change ignores both listeners when running in AOT mode

Closes gh-1561

### DIFF
--- a/spring-aot/src/main/java/org/springframework/aot/factories/IgnoredFactoriesCodeContributor.java
+++ b/spring-aot/src/main/java/org/springframework/aot/factories/IgnoredFactoriesCodeContributor.java
@@ -17,7 +17,8 @@
 package org.springframework.aot.factories;
 
 import java.util.Arrays;
-import java.util.List;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.function.Predicate;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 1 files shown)


spring-aot/src/main/java/org/springframework/aot/factories/IgnoredFactoriesCodeContributor.java
```diff
@@ -52,12 +53,21 @@
 			"org.springframework.boot.test.autoconfigure.filter.TypeExcludeFiltersContextCustomizerFactory",
 			"org.springframework.boot.test.context.ImportsContextCustomizerFactory");
 
+	private static final Predicate<SpringFactory> IGNORED_TEST_EXECUTION_LISTENERS = factoryEntry(
+			"org.springframework.test.context.TestExecutionListener",
```
sdeleuze, line 57: Could you elaborate on why `org.springframework.test.context.TestExecutionListener` exclusion is needed? Sounds like a general purpose one that could likely break regular tests if removed.

mhalbritter: The first parameter of `factoryEntry` is the key of the entry in `spring.factories`. This change only exclude the two mockito listeners, nothing else.

sdeleuze: Good point, that's great then.

sdeleuze (approved):

Issue #1538, opened 2022-03-22
GraalVM for Apple Silicon: `management.metrics.distribution` error


mraible:
I was motivated to try out [GraalVM for Apple Silicon](https://github.com/graalvm/graalvm-ce-dev-builds/releases/tag/22.1.0-dev-20220321_2332) after it was [announced today](https://twitter.com/fniephaus/status/1506192466705326082). 

How I installed it:

1. Downloaded and extracted [graalvm-ce-java11-darwin-aarch64-dev.tar.gz](https://github.com/graalvm/graalvm-ce-dev-builds/releases/download/22.1.0-dev-20220321_2332/graalvm-ce-java11-darwin-aarch64-dev.tar.gz)

2. Installed it as an SDKMAN candidate:

        mv ~/Downloads/graalvm-ce-java17-22.1.0-dev/Contents/Home ~/.sdkman/candidates/java/22.1.0.dev.r17-grl
        sudo xattr -r -d com.apple.quarantine ~/.sdkman/candidates/java/22.1.0.dev.r17-grl

3. Set SDKMAN to use it: `sdk default java 22.1.0.dev.r17-grl`

I used my [recent example with JHipster](https://developer.okta.com/blog/2022/03/03/spring-native-jhipster) on my M1 and ran:

```
./mvnw package -Pnative,prod -DskipTests
```

I was pleased to see it finished in 2m (vs 3m 10s on Intel). Unfortunately, it doesn't work if I try to start it:

```
$ ./target/native-executable
...
***************************
APPLICATION FAILED TO START
***************************

Description:

Failed to bind properties under 'management.metrics.distribution' to org.springframework.boot.actuate.autoconfigure.metrics.MetricsProperties$Distribution:

    Reason: java.lang.ClassNotFoundException: org.springframework.boot.actuate.autoconfigure.metrics.ServiceLevelObjectiveBoundary

Action:

Update your application's configuration
```

If I add `org.springframework.boot.actuate.autoconfigure.metrics.ServiceLevelObjectiveBoundary.class` to my type hints, it works.

```diff
--- a/src/main/java/com/auth0/flickr2/Flickr2App.java
+++ b/src/main/java/com/auth0/flickr2/Flickr2App.java
@@ -33,6 +33,7 @@ import tech.jhipster.config.JHipsterConstants;
         com.drew.metadata.exif.ExifThumbnailDirectory.class,
[...]

PR #1565, opened 2022-03-31
Explicity include ServiceLevelObjectiveBoundary

mhalbritter:
This triggers only if MetricsAutoConfiguration is active.

It looks like that this is a GraalVM 22.1 regression. This fix works with both 22.0 and 22.1.

Closes gh-1538

### DIFF
--- a/spring-native-configuration/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsEndpointAutoConfigurationHints.java
+++ b/spring-native-configuration/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsEndpointAutoConfigurationHints.java
@@ -19,8 +19,8 @@ package org.springframework.boot.actuate.autoconfigure.metrics;
 import org.springframework.boot.actuate.autoconfigure.metrics.export.properties.PropertiesConfigAdapter;
 import org.springframework.boot.actuate.autoconfigure.metrics.export.simple.SimplePropertiesConfigAdapter;
 import org.springframework.boot.actuate.metrics.MetricsEndpoint;
-import org.springframework.nativex.hint.TypeAccess;
 import org.springframework.nativex.hint.NativeHint;
+import org.springframework.nativex.hint.TypeAccess;
 import org.springframework.nativex.hint.TypeHint;
 import org.springframework.nativex.type.NativeConfiguration;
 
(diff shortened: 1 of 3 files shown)


spring-native-configuration/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsEndpointAutoConfigurationHints.java
```diff
@@ -19,18 +19,16 @@
...
 import org.springframework.nativex.hint.TypeHint;
 import org.springframework.nativex.type.NativeConfiguration;
 
 // Hitting /metrics endpoint - needs tests
 @NativeHint(trigger = MetricsEndpointAutoConfiguration.class, types = {
 	@TypeHint(types = {
 		PropertiesConfigAdapter.class,
-		SimplePropertiesConfigAdapter.class,
-		org.springframework.boot.actuate.autoconfigure.metrics.ServiceLevelObjectiveBoundary.class,
-		org.springframework.boot.actuate.autoconfigure.metrics.ServiceLevelObjectiveBoundary[].class,
```
sdeleuze, line 33: Since `ServiceLevelObjectiveBoundary.class` and `ServiceLevelObjectiveBoundary[].class` require distinct configuration (at least it was the case until now), should we keep `ServiceLevelObjectiveBoundary[].class` here since it was needed for the endpoint as far as I remember?

mhalbritter: The `ServiceLevelObjectiveBoundary[]` entries are still in the generated `reflect-config.json` file:

```
  {
    "name": "org.springframework.boot.actuate.autoconfigure.metrics.ServiceLevelObjectiveBoundary"
  },
  {
    "name": "org.springframework.boot.actuate.autoconfigure.metrics.ServiceLevelObjectiveBoundary[]"
  }
```

I'm not sure where the array one is coming from. But of course we can explicitly include them, just to be sure.

sdeleuze: Ok likely generated by the improved inference mechanism I guess.

Issue #1580, opened 2022-04-13
AOP do not support '*' on class field with execution pointcut designator 


Sochene:
When I test samples/class-proxies-aop, modify the  LoggableAspect.java GraalvmDemoApplication,java
`
        @Pointcut("execution(* com.example.graalvmdemo.rest.Person*.demo(..))")
`

```
@AotProxyHint(targetClass=com.example.graalvmdemo.rest.PersonController.class,proxyFeatures = ProxyBits.IS_STATIC)
@AotProxyHint(targetClass=com.example.graalvmdemo.rest.PersonController2.class,proxyFeatures = ProxyBits.IS_STATIC)
```

it build success, but run failed with logs below
```
2022-04-13 19:31:50.574 ERROR 2674732 --- [           main] o.s.boot.SpringApplication               : Application run failed

org.springframework.context.ApplicationContextException: Unable to start web server; nested exception is org.springframework.boot.web.server.WebServerException: Unable to start embedded Tomcat
	at org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext.onRefresh(ServletWebServerApplicationContext.java:163) ~[na:na]
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:577) ~[na:na]
	at org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext.refresh(ServletWebServerApplicationContext.java:145) ~[na:na]
	at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:740) ~[class-proxies-aop:2.6.6]
	at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:415) ~[class-proxies-aop:2.6.6]
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:303) ~[class-proxies-aop:2.6.6]
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1312) ~[class-proxies-aop:2.6.6]
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1301) ~[class-proxies-aop:2.6.6]
	at com.example.graalvmdemo.GraalvmDemoApplication.main(GraalvmDemoApplication.java:13) ~[class-proxies-aop:0.0.1-SNAPSHOT]
Caused by: org.springframework.boot.web.server.WebServerException: Unable to start embedded Tomcat
[...]

PR #1601, opened 2022-05-03
Fix Spring AOP AspectJ PCD #1580

Sochene:
fix [issues/1580](https://github.com/spring-projects-experimental/spring-native/issues/1580)

Because of RequestContextFilter lose destroy() method, ClassUtils.getMostSpecificMethod return Filter#destroy.


sdeleuze:
Thanks for this PR, could you please modify one of the samples to test this use case?

Sochene:
> Thanks for this PR, could you please modify one of the samples to test this use case?

ok

Issue #459, opened 2021-01-29
Unable to generate proxy for class that implements multiple interfaces.


mminella:
In Spring Batch, there are two custom scopes `StepScope` and `JobScope`. In each of these scenarios, the bean's definition is proxied so that the underlying bean can be created later during execution. This functionality allows for things like the ability to inject job parameters into Spring Batch components.

However, the proxies being created when using a native image are based on the return type of a method defined with the `@Bean` annotation. If I have a class that implements multiple interfaces that must be exposed via the poxy, there is no way currently to generate that.

An example of the issue can be found here: https://github.com/mminella/spring-graalvm-native/tree/batch_io. In that example, in the `BatchConfiguration` class, there is a bean defined as:

```java 
@StepScope
@Bean
public ItemReaderListener reader() {
   ...
}
```

`ItemReaderListener` is a class that implements the `ItemReader`, `ItemStream`, and the `StepExecutionListener` interfaces. Spring Batch provides a composite interface for `ItemReader` and `ItemStream` so the method can be refactored to return `ItemStreamReader`. However, the proxy created based on that definition cannot be cast to a `StepExecutionListener`.

sdeleuze:
Any thoughts on that one @aclement?

aclement:
I wrestled with this one (with Michael) last week. My initial reaction was to make the proxy creation we have in spring a little smarter, but I haven't dug into it.  Right now we have that patch which says "if not native-image do the right thing, otherwise just give them a jdkproxy", I wondered about that native-image clause, making it do a little more analysis (certainly produce more diagnostics because when a bad situation arises it causes chaos downstream trying to debug).

sdeleuze:
So we should have something more clever that tries to create a JDK proxy configuration for the various interfaces implemented by the related class?

mminella:
The creating a JDK proxy configuration for the various interfaces is what I thought I was attempting to do in my `BatchScopeComponentProcessor` here: https://github.com/mminella/spring-graalvm-native/blob/batch_io/spring-native-configuration/src/main/java/org/springframework/batch/core/configuration/annotation/BatchScopeComponentProcessor.java but to no avail.

aclement:
> So we should have something more clever that tries to create a JDK proxy configuration for the various interfaces implemented by the related class?

[...]

PR #1618, opened 2022-05-20
Update batch-io sample with FlatFileItemReader

fmbenhassine:
This commit updates the `batch-io` sample to make the item reader reads data from a flat file (disk IO), as the sample name implies.

However, with this change, the sample fails with the same issue reported in #459 . Any idea why the sample works when the reader extends `AbstractItemCountingItemStreamItemReader` but not when it extends `FlatFileItemReader`?

### DIFF
--- a/samples/batch-io/src/main/java/com/example/batch/ItemReaderListener.java
+++ b/samples/batch-io/src/main/java/com/example/batch/ItemReaderListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021-2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


sdeleuze:
Indeed I see the error:
```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'job': Unsatisfied dependency expressed through method 'job' parameter 0; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'step1': Unsatisfied dependency expressed through method 'step1' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'reader': Invocation of init method failed; nested exception is org.springframework.beans.factory.support.ScopeNotActiveException: Error creating bean with name 'scopedTarget.reader': Scope 'step' is not active for the current thread; consider defining a scoped proxy for this bean if you intend to refer to it from a singleton; nested exception is java.lang.IllegalStateException: No context holder available for step scope
	at org.springframework.aot.beans.factory.InjectedConstructionResolver.resolve(InjectedConstructionResolver.java:88) ~[na:na]
	at org.springframework.aot.beans.factory.InjectedElementResolver.resolve(InjectedElementResolver.java:35) ~[batch-io:0.12.1-SNAPSHOT]
	at org.springframework.aot.beans.factory.InjectedElementResolver.create(InjectedElementResolver.java:66) ~[batch-io:0.12.1-SNAPSHOT]
	at org.springframework.aot.beans.factory.BeanDefinitionRegistrar$BeanInstanceContext.create(BeanDefinitionRegistrar.java:211) ~[na:na]
...
Caused by: java.lang.IllegalStateException: No context holder available for step scope
	at org.springframework.batch.core.scope.StepScope.getContext(StepScope.java:167) ~[batch-io:4.3.6]
	at org.springframework.batch.core.scope.StepScope.get(StepScope.java:99) ~[batch-io:4.3.6]
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:371) ~[batch-io:5.3.21]
	... 61 common frames omitted
```
Since we focus mostly on Spring Boor 3 these days, I think I am going to just apply it on the https://github.com/spring-projects-experimental/spring-native/tree/sb-3.0.x branch unless @aclement see something obvious we can fix.

sdeleuze:
Hum, your PR seems to break the AOT mode on the JVM as well (tested with `./build.sh -a` on both `main` and `sb-3.0.x` branches). Pinging @snicoll in case this could be something we don't support yet.

snicoll:
Superseded by the AOT smoke tests.

PR #1670, opened 2022-07-13
Support lambda class serialization via hints

hanakeichen:
Currently lambdas serialization https://github.com/oracle/graal/commit/b455f23b624a1af149c7d5769443a4775bedd087 is supported in GraalVM native-image. This commit add lambdaCapturingTypes and lambdaCapturingTypeNames to conveniently add capturing classes to serialization configuration.

Example:

```
public class LambdaClassSerialization {
    private static void serialize(ByteArrayOutputStream byteArrayOutputStream, Serializable serializableObject) throws IOException {
        ObjectOutputStream objectOutputStream = new ObjectOutputStream(byteArrayOutputStream);
        objectOutputStream.writeObject(serializableObject);
        objectOutputStream.close();
    }

    private static Object deserialize(ByteArrayOutputStream byteArrayOutputStream) throws IOException, ClassNotFoundException {
        ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
        ObjectInputStream objectInputStream = new ObjectInputStream(byteArrayInputStream);
        return objectInputStream.readObject();
    }

    public static void testLambdaClassSerialization() throws Exception {
        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
        Function<Integer, String> function = (Function<Integer, String> & Serializable) (x) -> "Hello";

        serialize(byteArrayOutputStream, (Serializable) function);
        (Function<Integer, String>) deserialize(byteArrayOutputStream);
    }
}

// serialization-config.json output :
// 
// {
//   "types":[
//   ],
//   "lambdaCapturingTypes":[
//     {
//       "name":"com.xxx.LambdaClassSerialization"
//     }
//   ]
// }
@SerializationHint(lambdaCapturingTypes = {LambdaClassSerialization.class})
@SpringBootApplication
public class LambdaClassSerializationSpringApplication {

	public static void main(String[] args) throws Exception {
		SpringApplication.run(LambdaClassSerializationSpringApplication.class, args);
	}

}
```

### DIFF
--- a/spring-aot/src/json-shade/java/org/springframework/nativex/json/JSONArray.java
+++ b/spring-aot/src/json-shade/java/org/springframework/nativex/json/JSONArray.java
@@ -44,7 +44,7 @@ import java.util.List;
  * overridable methods is not specified. See <i>Effective Java</i> Item 17, "Design and
  * Document or inheritance or else prohibit it" for further information.
  */
-public class JSONArray {
+public class JSONArray implements JSONValue {
 
 	private final List<Object> values;
 
--- a/spring-aot/src/json-shade/java/org/springframework/nativex/json/JSONObject.java
+++ b/spring-aot/src/json-shade/java/org/springframework/nativex/json/JSONObject.java
@@ -72,7 +72,7 @@ import java.util.Map;
  * overrideable methods is not specified. See <i>Effective Java</i> Item 17, "Design and
  * Document or inheritance or else prohibit it" for further information.
  */
-public class JSONObject {
+public class JSONObject implements JSONValue {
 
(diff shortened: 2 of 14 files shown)


sdeleuze:
Since Spring Boot 3.0 GA has been released, Spring Native is not actively developed, so I will close this PR unmerged. I don't think we support lambda class serialization in Spring Framework 6, and not sure yet if we should support it or not yet. We should be sure to have enough use cases if we want to support it, so if you create a PR/issue on Spring Framework side, make sure to provide more details on your use cases.

PR #1673, opened 2022-07-18
maybe we could add primary-classes option for spring-aot generation

wineway:
***this is a draft pr***

When we start the spring application:
```
-               SpringApplication.run(new Class[] {
-                               Xxxx1Configuration.class,
-                               Xxxx2Configuration.class,
-                               Application.class
-               }, args);
```

spring native creates context using application class and ignores other classes, this will cause some registration of beans to be missing, maybe we can provide the manual configurable option to solve this problem.

Signed-off-by: wineway <wangyuweihx@gmail.com>


sdeleuze:
Thanks for your proposal but Spring Native is now in maintenance mode as we focus more our bandwidth on Spring Framework 6 and Spring Boot 3 native support, so I will close this PR has it is not fixing a critical bug as far as I can say.

Issue #1532, opened 2022-03-13
Restore support for Spring Cloud Stream


ivangfr:
Hi, I've upgraded my project [ivangfr/spring-cloud-stream-event-routing-cloudevents](https://github.com/ivangfr/spring-cloud-stream-event-routing-cloudevents) to **spring-boot** `2.6.4` and **spring-native** `0.11.3`.

The project has two apps: `producer-service` and `consumer-service`.

The Docker native image is built successfully. However, the following exceptions are thrown

- `producer-service` the following exception is thrown when a `news` or an `alert` is submitted
  ```
  ERROR 1 --- [ctor-http-nio-3] a.w.r.e.AbstractErrorWebExceptionHandler : [7f7c07d9-3]  500 Server Error for HTTP POST "/api/alerts/weather"
  
  java.lang.IllegalArgumentException: Attribute 'value' not found in attributes for annotation [org.springframework.context.annotation.Role]
  	at org.springframework.util.Assert.notNull(Assert.java:219) ~[na:na]
  	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException:
  Error has been observed at the following site(s):
  	*__checkpoint ? org.springframework.boot.actuate.metrics.web.reactive.server.MetricsWebFilter [DefaultWebFilterChain]
  	*__checkpoint ? HTTP POST "/api/alerts/weather" [ExceptionHandlingWebHandler]
  Original Stack Trace:
  		at org.springframework.util.Assert.notNull(Assert.java:219) ~[na:na]
  		at org.springframework.core.annotation.AnnotationAttributes.assertAttributePresence(AnnotationAttributes.java:366) ~[na:na]
  		at org.springframework.core.annotation.AnnotationAttributes.getRequiredAttribute(AnnotationAttributes.java:353) ~[na:na]
  		at org.springframework.core.annotation.AnnotationAttributes.getNumber(AnnotationAttributes.java:222) ~[na:na]
  		at org.springframework.context.annotation.AnnotationConfigUtils.processCommonDefinitionAnnotations(AnnotationConfigUtils.java:259) ~[na:na]
  		at org.springframework.context.annotation.AnnotationConfigUtils.processCommonDefinitionAnnotations(AnnotationConfigUtils.java:234) ~[na:na]
  		at org.springframework.context.annotation.AnnotatedBeanDefinitionReader.doRegisterBean(AnnotatedBeanDefinitionReader.java:263) ~[na:na]
[...]

PR #1677, opened 2022-07-29
Adds back in cloud-stream-rabbit sample

onobc:
This adds back in the SCSt Rabbit binder sample. 

* Current state of SCSt wrt to AOT/Native is that both Kafka and Rabbit binder samples work in AOT mode but not yet in Native mode. 
* Once this is merged these 2 samples can be moved into the smoke test repo. 

@mhalbritter do you mind merging this? 

Contributes toward #1532


mhalbritter:
Thanks!

PR #1690, opened 2022-08-13
Move cloud-stream-kafka/rabbit samples to Spring AOT smoke tests

onobc:
This proposal removes all `cloud-stream-[kafka|rabbit]` samples and leaves behind a README.md pointing to their new home.


sdeleuze:
I think the target is different (Spring Boot 2.x versus 3.x) so for now I think I would prefer to keep the samples on Spring Native side. Once Spring Boot 3 GA is released, I think we will do some global updates to Spring Native and its samples to redirect to Spring Boot 3, Spring AOT smoke tests and related  documentation. As a consequence, I don't merge this PR. Feel free to let me know if that does not make sense from your POV.

sdeleuze:
Sorry I miss it was on `sb-3.0.x` branch.

